PR #3, opened 2019-02-28
Merge upstream 2019-02-27

dongsupark:
To make the build chain rely on perl 5.26.2 instead of perl 5.24, we should merge upstream.

(Draft PR. Let's not merge yet.)


dongsupark:
Not needed any more.

PR #33, opened 2020-03-19
acct-*: add acct packages for openssh 8.1 for edge

dongsupark:
As dependencies needed by net-misc/openssh 8.1, add `acct-group/sshd`, `acct-user/sshd`, `virtual/libgcrypt`.

To do that, we need to add new categories, `acct-group` and `acct-user`.
Also we need to update eclasses like `user-info`, `user`, `acct-group`, `acct-user`.


dongsupark:
This PR is needed for updating further packages like qemu.
Merge.

PR #59, opened 2020-04-29
dev-util/pkgconfig: bump to 0.29.2 for PKG_CHECK_MODULES_STATIC for alpha

dongsupark:
pkg-config 0.29 or newer introduced a macro `PKG_CHECK_MODULES_STATIC`,
which is needed by ebuilds like dev-util/strace >= 5.5.
So we should update pkg-config to 0.29.2, to avoid such issues.
(see https://gitlab.freedesktop.org/pkg-config/pkg-config/commit/0efb668bd5990f52a6a93b2cde037185edbe11b0)

```
***** autoconf *****
***** PWD: /build/amd64-usr/var/tmp/portage/dev-util/strace-5.6/work/strace-5.6
***** autoconf --force -I /build/amd64-usr/usr/share/aclocal

configure:17585: error: possibly undefined macro: PKG_CHECK_MODULES_STATIC
```

## How to use

```
emerge-amd64-usr dev-util/pkgconfig
```

Note, it is not a blocker for strace PRs, which we can simply work around by removing a third-party patch that calls PKG_CHECK_MODULES_STATIC.


dongsupark:
Gentle ping.
Please review it, so we see the same issue again in the future. :-)

PR #66, opened 2020-05-07
Update libraries related to libselinux 3.0

dongsupark:
To be able to update libselinux to 3.0, we need to update the following packages:

* sys-libs/libsepol

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/347, https://github.com/flatcar-linux/scripts/pull/66.


dongsupark:
Please see https://github.com/kinvolk/portage-stable/pull/172 instead

PR #79, opened 2020-06-22
dev-libs/libgit2: import libgit2 for alpha

dongsupark:
As `dev-lang/rust` >= 1.44.0 requires `dev-libs/libgit2`, we need to import libgit2 from upstream Gentoo.

As `dev-libs/libgit2` inherits `cmake.eclass`, we need to import it from upstream Gentoo.

## How to use

```
emerge-amd64-usr dev-libs/libgit2
```


dongsupark:
cherry-picked to edge

PR #49, opened 2020-04-22
dev-vcs/git: bump to 2.26.2 for alpha

dongsupark:
Update git to 2.26.2, to address the following security issues:

* [CVE-2020-5260](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-5260)
* [CVE-2020-11008](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11008)

## How to use

```
emerge-amd64-usr dev-vcs/git
```

PR #83, opened 2020-06-29
dev-vcs/git: bump to 2.26.2 for stable

dongsupark:
Update git to 2.26.2, to address the following security issues:

* [CVE-2020-5260](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-5260)
* [CVE-2020-11008](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11008)

## How to use

```
emerge-amd64-usr dev-vcs/git
```

*Note*: it is basically the same update as https://github.com/flatcar-linux/portage-stable/pull/49 and https://github.com/flatcar-linux/portage-stable/pull/50, but it is meant for stable and beta.
Since the upstream gentoo mirror http://distfiles.gentoo.org/distfiles/ removed a tarball http://distfiles.gentoo.org/distfiles/git-2.24.1.tar.xz , we are not able to rely on git 2.24.1 any more. It is also not possible to fetch the tarball from https://www.kernel.org/ via https, because of another [issue](https://github.com/flatcar-linux/Flatcar/issues/149). That's why it is necessary to update git to the latest version.


dongsupark:
cherry-picked to beta

PR #84, opened 2020-07-03
eclass: sync linux-info.eclass to fix Kernel build issues

dongsupark:
Kernel 5.8-rc2 or newer added a commit [f2f02ebd8f38](https://github.com/torvalds/linux/commit/f2f02ebd8f3833626642688b2d2c6a7b3c141fa9)
("kbuild: improve cc-option to clean up all temporary files") , which causes make commands to create an object output directory.
Apparently harmless.
The commit was backported to stable Kernels, 4.19.131, 5.4.50, 5.7.7.

In Flatcar and Gentoo, however, `getfilevar()` in `linux-info.eclass` runs a make command, which creates `${M}`, an object output directory, with a root account.
As Gentoo sandbox creates everything as non-root account, the subsequent steps like `src_unpack` fail to touch the same directory.

Upstream Gentoo already has a fix [5a3acd443c33](https://github.com/gentoo/gentoo/commit/5a3acd443c3347568d14d000776f46177317744a) ("linux-info.eclass: Pass M=${T} to the Linux Makefile unconditionally.").
See also https://bugs.gentoo.org/729178 .

So simply sync `linux-info.eclass` with Gentoo.

## How to use

```
emerge-amd64-usr sys-kernel/coreos-modules
```

### DIFF
--- a/eclass/linux-info.eclass
+++ b/eclass/linux-info.eclass
@@ -1,4 +1,4 @@
-# Copyright 1999-2018 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: linux-info.eclass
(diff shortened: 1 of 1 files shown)


dongsupark:
Tested builds with new Kernels like 5.4.50.
Tested builds with an older Kernel like 4.19.130.
Have seen no regression.

dongsupark:
Thanks for reviews!

dongsupark:
cherry-picked to beta, alpha, and edge.

PR #95, opened 2020-08-07
 Update packages for openssl update

pothos:
See https://github.com/flatcar-linux/coreos-overlay/pull/514

(_app-crypt/mit-krb5: Fix Python version_ is a downstream modification but too trivial for coreos-overlay and nothing we want to keep)


app-crypt/mit-krb5/mit-krb5-1.18.2-r1.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{6,7} )
```
t-lo, line 6: Why not just `python3_7`? The original recipe does not list `python3_6` as compatible.

pothos: We have ebuilds for 3.6 and 3.7 but only 3.6 is built at the moment.

t-lo: Oh, okay, makes sense then, thanks for elaborating. The tests show it builds and it works with `python3_6`, so we're good to go.

pothos:
Picked for flatcar-2592

PR #98, opened 2020-08-25
acct-group: add group "render" for udev

t-lo:
### cherry-pick to: `flatcar-2605`

This is a requirement for flatcar-linux/coreos-overlay#543; please see that PR for description & instructions.

Addresses flatcar-linux/Flatcar#169 in the OS image.


t-lo:
Not needed, see https://github.com/flatcar-linux/coreos-overlay/pull/543#discussion_r476499498.

PR #99, opened 2020-08-26
sys-fs/mdadm: Drop it, temporarily

krnowak:
We are going to have mdadm in coreos-overlay. It will contain an extra
patch, so it can't be a part of portage-stable.

It should be merged together with flatcar-linux/coreos-overlay#546.


t-lo (approved):
LGTM, please merge **after** https://github.com/flatcar-linux/coreos-overlay/pull/546.
Please also cherry-pick to `flatcar-2605`.

krnowak:
Cherry-pick done.

PR #100, opened 2020-08-27
sys-apps/iproute2: bump to 5.8.0, add ELF support

t-lo:
# Update `iproute2` to 5.8.0, add ELF support

This PR adds Gentoo upstream's `iproute2-5.8.0.ebuild` recipe and enables ELF support to e.g. allow `tc` to handle BPF filters. 

**NOTE** that the current release tarball of `iproute2-5.8.0` (https://mirrors.edge.kernel.org/pub/linux/utils/net/iproute2/) builds version `5.7.0.77` (defined in `include/version.h` of the sources).

# How to use

1. Fetch the Flatcar SDK for Alpha-2605.0.0: `cork create --manifest-branch refs/tags/v2605.0.0`
2. In the Flatcar SDK, check out branch `t-lo/iproute2-5.8.0-with-elf-support` in `src/third_party/portage-stable`.
3. Build an image:
   ```
   $ cork enter
   $ ./setup_board --board=amd64-usr
   $ ./build_packages --board=amd64-usr
   $ ./build_image --board=amd64-usr
   $ ./image_to_vm.sh --from=../build/images/amd64-usr/developer-latest --board=amd64-usr
   $ ../build/images/amd64-usr/developer-latest/flatcar_production_qemu.sh
   ```
4. ssh into the image, check `tc` output as described in https://github.com/flatcar-linux/Flatcar/issues/172#issuecomment-680073152:
   ```
   ssh core@localhost -p2222
   $ tc filter add dev bond0 egress bpf da obj mybpf.elf sec filter_egress
   Error opening object mybpf.elf: No such file or directory
   Cannot initialize ELF context!
   Unable to load program
   ```
   (the error message is expected because `mybpf.elf` is missing. The message would be different if ELF support was missing)

# Testing done

Ran the above steps.

### cherry-pick to `flatcar-2605` if integration test suite succeeds. 

Implements https://github.com/flatcar-linux/Flatcar/issues/172


profiles/base/package.use
```diff
@@ -31,3 +31,5 @@ app-admin/sudo -ldap
 # Bug #295615 - Disable cxx by default, in order to avoid a die
 # in pkg_setup with default USE settings (cxx requires sasl).
 net-nds/openldap -cxx
+
+sys-apps/iproute2 elf
```
dongsupark, line 35: It is ok-ish, but,
if we add Flatcar-specific changes in portage-stable, then it could be overwritten by upstream changes, next time when we sync with Gentoo for example.
That's why we usually gather all the Flatcar-specific profile changes in `coreos-overlay/profiles/`.

dongsupark (commented):
Looks good.

See below:

krnowak:
This directory does not seem to be in sync with upstream Gentoo. There are some old ebuilds too. Could you do an update by basically doing `rm -rf sys-apps/iproute2; cp -a ${GENTOO_REPO}/sys-apps/iproute2 sys-apps/iproute2` instead?

profiles/base/package.use
t-lo: I'd rather not put USE flags for packages in `portage-stable` into the profile of `coreos-overlay`. Profiles exist per overlay and should cover packages in the respective overlay.
You're rightfully pointing out that we need to pay attention when merging Gentoo changes though. We'll need to tread carefully there.

t-lo:
> This directory does not seem to be in sync with upstream Gentoo. There are some old ebuilds too. Could you do an update by basically doing `rm -rf sys-apps/iproute2; cp -a ${GENTOO_REPO}/sys-apps/iproute2 sys-apps/iproute2` instead?

I agree with cleaning up but would first like to get Beta 2605 out, i.e. separate concerns. Leaving the old recipes there for the time being enables us to go back to a previous version if the upgrade breaks tests.

krnowak:
> > This directory does not seem to be in sync with upstream Gentoo. There are some old ebuilds too. Could you do an update by basically doing `rm -rf sys-apps/iproute2; cp -a ${GENTOO_REPO}/sys-apps/iproute2 sys-apps/iproute2` instead?
> 
> I agree with cleaning up but would first like to get Beta 2605 out, i.e. separate concerns. Leaving the old recipes there for the time being enables us to go back to a previous version if the upgrade breaks tests.

I mean - if the upgrade breaks tests then you can do `git revert`. But anyway, let's get this in and we'll get there with cleaning up soon.

t-lo:
@krnowak @dongsupark 
icanhas_approval

profiles/base/package.use
dongsupark: I am not sure I completely agree, but anyway let's merge it for now.

dongsupark (approved):

t-lo:
Merged to main and [cherry-picked for `flatcar-2605`](https://github.com/flatcar-linux/portage-stable/commit/6c860ed5b24198439d122d79ad18520d722d3022).

PR #103, opened 2020-09-09
.github: Improve body text of PR for virtual/rust

dongsupark:
Improve body text of each PR for `virtual/rust`, by mentioning that it should be merged together with its paired PR in coreos-overlay.

Explicitly name `virtual/rust` instead of `Cargo`, because there is no more ebuild for `cargo`.
Rename the dispatched event-type name to `rust-pull-request-main`, as `cargo` has already disappeared.

Make the repository-dispatch action receive additional client-payload with a field `coreos-overlay-pull-request-number` sent by the corresponding PR in coreos-overlay.
The PR number is then used for adding a link in the body text, for pointing back to the PR in coreos-overlay.

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/575.

### DIFF
--- a/.github/workflows/rust-pull-request-main.yml
+++ b/.github/workflows/rust-pull-request-main.yml
@@ -1,6 +1,6 @@
 on:
   repository_dispatch:
-    types: [cargo-pull-request-main]
+    types: [rust-pull-request-main]
 
 jobs:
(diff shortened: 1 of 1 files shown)


marga-kinvolk (approved):
Thanks!

PR #105, opened 2020-09-18
net-dns/bind-tools: Update to version 9.11.23

marga-kinvolk:
Version 9.11.22 is gone and 2512 doesn't build anymore otherwise.


t-lo (approved):
Thank you for the prompt action! LGTM.

PR #108, opened 2020-09-23
dev-db/sqlite: bring back sqlite 3.31.1 for SDK

dongsupark:
To make the SDK build work again, we need to bring back dev-db/sqlite
3.31.1 in portage-stable. It is not enough to have it in coreos-overlay.

This reverts commit 4a7a4e3d272812963c3cd21431d1849ca9df11e4.


marga-kinvolk (approved):
LGTM

PR #110, opened 2020-09-25
sys-apps/makedev: bring back the package for bootstrap_sdk

dongsupark:
The `bootstrap_sdk` stage still requires `sys-apps/makedev` to be
available, as listed in [`profiles/default/linux/packages.build`](https://github.com/flatcar-linux/portage-stable/blob/ce68a65de4f24a561d00e07d23fa536b6b3728c2/profiles/default/linux/packages.build#L21).
We need to bring it back to make the SDK build work again.

This reverts commit df8159f565972eb31455ff5e4cbfba8c4a12bb52.

## Testing done

TBD


marga-kinvolk (approved):
LGTM

PR #112, opened 2020-09-30
Update dracut

krnowak:
This PR is a part of systemd update.

It updates dracut, so we can build a latest version, that is patched by gentoo to avoid using the deprecated values in `StandardOutput=` and `StandardError=` options in some systemd service files. This is only to avoid some warnings.

It also synces the systemd.eclass with upstream (mostly some comment changes).

It pulls the optfeature.eclass from upstream, since `optfeature` stopped being a part of eutils eclass for EAPI 7 (which dracut now uses).

The test build is http://localhost:9091/job/os/job/manifest/1138/


pothos:
The tests have a failure for `initrd-setup-root.service` and report `systemd-tmpfiles[633]: baselayout-home.conf:2: Failed to resolve user 'core': No such process`

krnowak:
This is fixed in bootengine.

dongsupark (approved):
Looks good, if CI passed.

PR #116, opened 2020-10-21
app-arch/tar: Sync package with upstream

sayanchowdhury:
# app-arch/tar: Sync package with upstream

# How to use

```
emerge-amd64-usr tar
```

# Testing done

- I've tested the changes locally.
- Triggered a Jenkins build for the same - http://localhost:9091/job/os/job/manifest/1251/cldsv/. Retriggered qemu_uefi.


dongsupark (approved):
Looks good. Thanks!

PR #120, opened 2020-10-26
app-arch/zstd: Import from Gentoo

krnowak:
An updated systemd will need it.

Just did a sync of systemd to latest ebuild from Gentoo and there is a new USE flag, enabled by default, `zstd`. This pulls a dep on the `app-arch/zstd` package, which I'm adding here.

http://localhost:9091/job/os/job/manifest/1300/


vbatts:
zstd is going to be nice to have around

dongsupark (approved):
Looks good if CI passed.

Nit: it would be nice to have `zstd` in the PR title.

krnowak:
> Looks good if CI passed.
> 
> Nit: it would be nice to have `zstd` in the PR title.

Oooops, I meant to have it in the title. Fixed.

krnowak:
Test build passed, merging.

PR #124, opened 2020-11-13
net-fs/cifs-utils: update to 6.11

t-lo:
Update net-fs/cifs-utils to Gentoo upstream 6.11.

# How to use / testing done
```
emerge-amd64-usr cifs-utils
```


dongsupark (approved):
Strange, it started pulling in `dev-python/docutils`.
Anyway we need it for security updates, so ok for now.

t-lo:
dev-python/docutils is a build time dependency (`BDEPEND`) which is used for generating man pages - it should not end up in the image.

PR #131, opened 2020-11-30
dev-util/dwarves: update to 1.19 for CONFIG_DEBUG_INFO_BTF

dongsupark:
It is already possible to enable kernel config `CONFIG_DEBUG_INFO_BTF` using dwarves 1.18 included in the current Flatcar SDK, as long as its arch is amd64.

However, Kernel build fails in case of arm64, when Kernel version is >= 5.9 and dwarves version is <= 1.18, like the following:

```
+ pahole -J .tmp_vmlinux.btf
PAHOLE: Error: Found symbol of zero size when encoding btf
(sym: '__kvm_nvhe_arm64_ssbd_callback_required', cu: '../source/arch/arm64/kernel/cpu_errata.c').
PAHOLE: Error: Use '-j' or '--force' to ignore such symbols and force emit the btf.
../source/scripts/link-vmlinux.sh: line 141: 1929102 Segmentation fault
```

The bug was fixed via https://git.kernel.org/pub/scm/devel/pahole/pahole.git/commit/?id=2e719cca6672, ("btf_encoder: revamp how per-CPU variables are encoded").
The fix was first included in dwarves [1.19](https://git.kernel.org/pub/scm/devel/pahole/pahole.git/tag/?h=v1.19).

Thus we need to get dwarves 1.19 included in Flatcar SDK, so that the next Alpha Kernels could have `CONFIG_DEBUG_INFO_BTF` enabled.

## How to use

```
emerge dev-util/dwarves
```

## Testing done

With dwarves 1.19 installed in the SDK, Kernel 5.9 build works for both amd64 and arm64.


marga-kinvolk (commented):
Is there a reason to keep the old versions around as well?

dongsupark:
> Is there a reason to keep the old versions around as well?

No.
In this case, I have simply run `update_ebuilds dev-util/dwarves`, which then automatically got portage-stable synced with upstream Gentoo.
That's the only reason why portage-stable repo keeps old ebuilds.
If the specific package in Gentoo did not have old ebuilds, then of course update_ebuilds would automatically remove them from portage-stable.

On the other hand, in case of coreos-overlay, we usually try to remove old ones, as we anyway have to manually maintain the repo.

marga-kinvolk (approved):
LGTM. Thanks for the explanation.

PR #149, opened 2021-02-22
net-misc/dhcpcd: delete dhcpcd

dongsupark:
As Flatcar relies on systemd-networkd for network configurations, it is not needed to keep dhcpcd in production images at all.

According to the commit https://github.com/kinvolk/coreos-overlay/commit/9be90f06e838 , it was added back in 2014 just because systemd-networkd was not mature enough.
That was already ~7 years ago, so we can safely assume that the issue had been already gone, so we can simply use systemd-networkd.

This PR should be merged together with https://github.com/kinvolk/coreos-overlay/pull/858.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2075/cldsv/ (part of a combined testing branch)


marga-kinvolk (approved):
LGTM

PR #150, opened 2021-02-25
dev-libs/icu: delete icu, update waf-utils.eclass

dongsupark:
To be able to use `EAPI=7` in `net-fs/samba`, we need to sync `waf-utils` with upstream Gentoo.

Now that `net-fs/samba` runs without spotlight backend, it is not needed to pull in unnecessary deps like `dev-libs/icu`.
Simply delete it.

This PR should be merged together with https://github.com/kinvolk/coreos-overlay/pull/864.

## Testing done

CI passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2106/


dongsupark:
CI passed

marga-kinvolk (approved):
LGTM

PR #151, opened 2021-03-02
sys-fs/multipath-tools: Sync with Gentoo upstream

sayanchowdhury:
# sys-fs/multipath-tools: Sync with Gentoo upstream

# How to use

```bash
emerge-amd64-usr sys-fs/multipath-tools
./build_image
```

# Testing done
Locally tested and Jenkins Build http://localhost:9091/job/os/job/manifest/2115/


marga-kinvolk:
The packages jobs corresponding to the Jenkins build are already gone... I guess we should keep them longer? Anyway, did you get to check the output of the job before it was gone?

dongsupark (approved):

PR #153, opened 2021-03-11
Boost 1.75.0 port

bonkko:
# Gentoo Boost Port
Port of Gentoo Boost 1.75.0
# How to use

see https://github.com/kinvolk/Flatcar/issues/362

# Testing done

see https://github.com/kinvolk/Flatcar/issues/362


marga-kinvolk:
Hi, Thanks for working on this.

I'm a bit confused about the refactor going on.  This package is in portage-stable. The packages in portage stable are mirrored as-is from Gentoo. If we need to make changes, then we move them to coreos-overlay, where we keep the packages that have Flatcar specific changes.

For very small things like specifying the python version so that it builds, it's arguably ok to make those changes here. But refactoring the ebuild is not a small change. And I'm not sure if it's even needed.

Could you please re-work this PR so that it includes one commit where you import from Gentoo verbatim and a second commit where do you only the absolutely necessary changes to make the package work?  Thanks.

bonkko:
Thanks for the response. 
The branch is portage-stable but in the system I use coreos-overlay. It is not very important for the compilation but it is important for your maintenance in Flatcar. I can use Gentoo verbatim for boost because I only change python version. 
I made a refactor of build-boost for the issue reported  in https://lists.boost.org/Archives/boost/2020/03/248336.php. I understand that it is a problem to maintain https://github.com/boostorg/build.git because a code review in the system is needed.

I will make fix for the 1.75.0 version aligned to Gentoo. The other version of the Gentoo boost library should work without problem.  I will remake my PR shortly in a proper way.

bonkko:
Hi,
With my last commit I made  a minimal change  from Gentoo vanilla(the branch is boost-1.75.0-port). I documented the activity in https://github.com/kinvolk/Flatcar/issues/362#issuecomment-796793902. Now the package use EAPI6 and python3_6. For  boost-1.75.0, I forced _-j1_  because with multiple jobs, on my machine, compilation fail . This not happen with boost-1.74.0-r2. 

Now the source code follow the Gentoo Community review. As noted in my tracking I'm open to take other step (at code level) for the port of this package.

marga-kinvolk:
Hi, thanks for the fix, but it's really hard to understand what's going on given the many different commits that you have here.  Please rebase your work so that you only have 2 commits in the PR: one commit with the code imported from Gentoo as is, followed by another commit with any necessary changes.

Please make sure those are the minimal amount of changes needed. For example, there's no reason to change the PV variable used in the ebuild file. That variable is filled in automatically by the build tools with the version that comes from the file name and there's no reason to touch it.  If you're having some trouble caused by that variable not being filled, it's a problem in the build toolchain, not in the ebuild file.

Thanks.

bonkko:
Hi, thanks for the feedback.
I made the two commit. The last is working with _stage4_ and a single _emerge_ but I think it is better to make a full test round. If you see any mistake in the last commit I will be happy to make the proper fix before doing my full test round.

Now the ebuilds are very similar to Gentoo.  I do not make test with mixed flag (for example may  be nice to remove the mpi because there is not any _sys-cluster_ in Flatcar or integrate Open MPI in the distro because currently the mpi flag is a dead flag in the ebuild).

bonkko:
Hi,
I made the round of test and all seems working. I think that now it is all ok but I'm welcome to improvement if needed.

dev-libs/boost/boost-1.75.0.ebuild
```diff
@@ -1,14 +1,14 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=6
```
dongsupark, line 4: From my understanding, you have to pin `MY_PV` to `1_75_0` and `MAJOR_V` to `1_75_0`, because you are using EAPI 6 here.
Is there a specific reason for doing that?
Would it be possible to set EAPI to 7, just like Gentoo does, so that you can use `ver_rs` and `ver_cut` below for `MY_PV` and `MAJOR_V`?

I am asking because in theory we should minimize the amount of code that is specific to Flatcar.
If it is really necessary to customize, then of course we could do so by moving the code to coreos-overlay.
I am just wondering if it is really necessary.

You wrote like `I have some little issue with the library that I think to solve in a short time.` in the [description](https://github.com/kinvolk/Flatcar/issues/362#issue-825432268).
I am not sure exactly what you mean.

dongsupark (commented):
Thanks for the PR.
It seems to work in general.
Please see the comments below:

dev-libs/boost/boost-1.75.0.ebuild
bonkko: Hi,
I  switched to EAPI 6 because you use it in your portage version. I did not see any EAPI 7 package in your portage. I can move to _Flatcar Alpha_  and put EAPI to 7 but I need to make test for it.

dongsupark: Right, the current ebuilds in Flatcar Alpha, `dev-libs/boost` 1.65.0 and `dev-util/boost-build` 1.65.0, have used EAPI 6.
That's simply because those ebuilds were written a few years ago, when Gentoo EAPI 7 was not widely used.
On the other hand, most ebuilds in the recent days rely on EAPI 7, following Gentoo standard.
For example, please see [app-arch/tar](https://github.com/kinvolk/portage-stable/blob/49516d78506a929c24de0b0546c643a5353367e7/app-arch/tar/tar-1.32.90.ebuild#L4).

As boost and boost-build need to be updated to 1.75, we should also change EAPI to 7 like Gentoo, unless there is a reason for not doing that.

bonkko: With a single test emerge finish the compilation with EAPI 7. I can switch to EAPI 7 if you prefer it but in my opinion it look a bit strange because in your portage directory I did not see any EAPI 7 packages. I only see EAPI=7 in coreos-overlay. 

In my interpretation portage-stable is Gentoo vanilla with minor modification  and with a stability in the system. On the other side coreos-overlay is a Flatcar specific overlay. I noted that you did not use EAPI 7 in portage-stable but you use it in coreos-overlay and I imagined that you did it for the stability of EAPI 7  with the Flatcar portage-stable. If it is not the case I can use EAPI 7 and make the test for the ebuild.

bonkko: Sorry , 
I read now post previous to my last comment (https://github.com/kinvolk/portage-stable/pull/153#discussion_r594262759). Thanks for the explanation , I will move to EAPI 7 and I will make the rebase after all my test round.

bonkko:
I made the  rebase with EAPI 7 and a test with stage4 and single emerge. The last test to do is emerge -e in the environment.

dongsupark:
Ok, the end result of the current PR looks good:
With EAPI 7, the functions `ver_rs` and `ver_cut` are being used.
Build seems to work as well.

Nitpick:

Your first commit `f54550f6597054f55974e60a05831a8f3ec0c5ce` should have `EAPI=7` for both  `boost-1.74.0-r2.ebuild` and `boost-1.75.0.ebuild`.
At the moment only the latter has EAPI 7, and `boost-1.74.0-r2.ebuild` still has EAPI 6 with hard-coded `MY_PV`/`MAJOR_V`.
As a result, the second commit `d9b8745eb75f31a5c016f4f9acf357ba3a2fdea1` still has unnecessary diffs in 1.74.0-r2.

Maybe other maintainers would need to have chances to look into it. 
Thanks! :-)

bonkko:
I'm sorry for the mistake with boost-1.74.0-r2.ebuild. Do you think it is preferable another PR  with the two strict necessary commit?

dongsupark:
Usually it is sufficient to force push to the existing PR, without creating another PR.
Anyway as I said, it was nitpick. ;-)

bonkko:
Noe it should be ok.

marga-kinvolk (commented):
Why is the metadata file in the Flatcar changes commit?

bonkko:
I'm sorry I made a mistake in the git add phase. I made the job by hand without the Flatcar scripts. For the next time it is better for me to follow the update howto.

bonkko:
Would you like another rebase?

marga-kinvolk:
Yes, please. We want the changes to be isolated to facilitate upgrades. If the metadata change is in the Flatcar commit, it will get in the way of future upgrades.

bonkko:
Ok, No problem I will do it.

bonkko:
Now there are only the changes with _python3_6_

marga-kinvolk:
Thanks, I'll do a CI run with this change.

bonkko:
Good.

marga-kinvolk:
CI passed.

bonkko:
Thanks. It is a nice notice. 

Do you think it is a good idea to start the list in the additional information section of the documentation of the PR (https://github.com/kinvolk/Flatcar/issues/362#issuecomment-796793902)?  This is not strictly related to the PR but I think it may be important for a long term port.

sayanchowdhury (approved):

PR #155, opened 2021-03-19
Add podman

jubalh:
# Add podman

Add podman and its dependencies (https://github.com/kinvolk/Flatcar/issues/112)

New packages:
* app-emulation/podman/podman
* app-crypt/gpgme
* app-emulation/conmon
* app-emulation/crun
* app-emulation/slirp4netns
* dev-go/go-md2man
* dev-libs/yajl
* net-libs/libslirp
* net-misc/cni-plugins
* sys-fs/fuse-common
* sys-fs/fuse-overlayfs
* sys-fs/fuse

Package updates:
* dev-libs/libassuan from 2.4.3 to 2.5.3, required by app-crypt/gpgme.

New eclass:
* eclass/go-module.eclass

I chose to add these to portage-stable instead of coreos-overlay, following the documentation that ebuilds that are imported from Gentoo and have few modifications should be put here.

Modifcations:
* disable tests for app-crypt/gpgme (includes omitting an instruction in `pkg_setup()`) since this tries to run gpg-agent.
* adjust `PYTHON_COMPAT` to `python3_7`

# How to use

Both dev-go/go-md2man and app-crypt/gpgme need to be installed in the SDK chroot.
I have the following packages unmasked in `chroot/build/amd64-usr/etc/portage/package.accept_keywords`:
```
app-emulation/podman ~amd64
app-emulation/conmon ~amd64
app-emulation/slirp4netns ~amd64
sys-fs/fuse-overlayfs ~amd64
app-emulation/crun ~amd64
app-crypt/gpgme ~amd64
dev-go/go-md2man ~amd64
sys-fs/fuse ~amd64
```

Do these need to be added to some profile?

For now when emerging podman the following is printed:

```
 * You need to create the following config files:
 * /etc/containers/registries.conf
 * /etc/containers/policy.json
 * To copy over default examples, use:
 * cp /etc/containers/registries.conf{.example,}
 * cp /etc/containers/policy.json{.example,}
 * 
 * For rootless operation, you need to configure subuid/subgid
 * for user running podman. In case subuid/subgid has only been
 * configured for root, run:
 * usermod --add-subuids 1065536-1131071 <user>
 * usermod --add-subgids 1065536-1131071 <user>
```

As does the upstream Gentoo ebuild. We could add these steps in our ebuild. However I checked a couple of our other ebuilds and couldn't find any that do similar things. So I'm unsure whether this should be left our or whether I need to do this in some other file/script.

# Testing done

TODO


pothos:
Hi,
that sounds good!
Are the `go-md2man` and `gpgme` packages build dependencies? Then I would expect them to be modeled like that in the ebuilds but a workaround could be to add them to `coreos-devel/sdk-depends` in coreos-overlay.
The accept keywords can go to `profiles/coreos/` in coreos-overlay.

It's good to try to minimize what we change but I think there will be changes needed and in the end some packages will go to coreos-overlay due to this. The main thing to be adjusted is to move the configuration files either to `/usr` (cf. how other ebuild files do this in coreos-overlay) or to use a hybrid approach where the template is in `/usr` and a systemd tmpfile directive copies/links this over to `/etc` if not present. Also, we would need to add some default the subuid file (again, not directly to `/etc` but in `/usr` and maybe using tmpfile to bring it to `/etc`).

dongsupark:
It pulls in quite a lot of new packages.

As for go-md2man, podman seems to pull in go-md2man as `DEPEND`. Since we use EAPI 7, we should use `BDEPEND` instead, to avoid installing unnecessary tools like that into the image. Though it is rather a Gentoo issue.
(See the [docs](https://mgorny.pl/articles/the-ultimate-guide-to-eapi-7.html#build-time-dependencies-split-into-depend-and-bdepend) for details)

We could either move podman to coreos-overlay for changing it to BDEPEND in its ebuild, or directly contribute to Gentoo to fix the issue. The latter is obviously desirable, but it could take time.

gpgme looks like an essential dependency of podman.
Assuming we would need rootless mode in the end, it is not simple to exclude fuse-overlayfs or slirp4netns.

So for now I just hope we could make some free space in the production image after rkt deprecation. :-)

jepio:
@jubalh thanks, this was good work but the world has changed and it will be implemented slightly differently (sysext).

PR #169, opened 2021-05-12
dev-util/cmake: update to 3.18.5

dongsupark:
Update `dev-util/cmake` to 3.18.5, as it is needed by `dev-lang/rust` 1.52.
Without that, building Rust 1.52 simply fails at the early stage.

Import a new package `dev-libs/jsoncpp`, as it is needed by `dev-util/cmake` 3.18.5.

It blocks https://github.com/kinvolk/coreos-overlay/pull/997 and https://github.com/kinvolk/portage-stable/pull/168 .

## How to use

```
emerge dev-util/cmake  # in the SDK
```

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2591/cldsv/


tormath1:
I don't get why upstream requires now `jsoncpp`. In 3.18.5, it seems we still have the ability to choose between the built-in one and the system one: https://gitlab.kitware.com/cmake/cmake/-/blob/v3.18.5/bootstrap#L618-619.
Built-in jsoncpp library is the `default` behavior but we do not override this flag (like in current [3.11.1](https://github.com/kinvolk/portage-stable/blob/main/dev-util/cmake/cmake-3.11.1.ebuild#L150)), it would mean that we pull `jsoncpp` for nothing ?
What do you think @dongsupark  ? :)

dongsupark:
Right. It looks strange.
Until cmake 3.16, its ebuild made use of the USE flag `system-jsoncpp`, so cmake pulled in jsoncpp only if `system-jsoncpp` was turned on.
However, since cmake 3.17, it is not the case any more. It simply pulls in `system-jsoncpp`, no matter whether the USE flag is enabled or not.
Not sure why, I tried to find out its reason, but could not.

We could either move cmake to coreos-overlay, to patch the ebuild file to disable `system-jsoncpp`.
Or we could directly fix cmake ebuilds in Gentoo.
Anyway I think it is not so critical issue at the moment, because make and jsoncpp are only installed in the SDK, which does not require tight space limitation.

tormath1:
> Anyway I think it is not so critical issue at the moment, because make and jsoncpp are only installed in the SDK, which does not require tight space limitation.

That's right, we can keep things in that way then !

tormath1 (approved):

PR #177, opened 2021-06-23
Unmasks for arm64 required by glibc-2.33 upgrade

t-lo:
This PR is required to [upgrade glibc to 2.33](https://github.com/kinvolk/coreos-overlay/pull/1067) in coreos-overlay.

To un-break the toolchain build for arm64 for the glibc-2.33 update, a few packages require un-masking for arm64.

This change unmasks arm64 for
  * app-misc/pax-utils
  * dev-libs/libev, libverto
  * virtual/acl

### DIFF
--- a/app-misc/pax-utils/pax-utils-1.1.6.ebuild
+++ b/app-misc/pax-utils/pax-utils-1.1.6.ebuild
@@ -14,7 +14,7 @@ SRC_URI="mirror://gentoo/${P}.tar.xz
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="alpha amd64 arm ~arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~s390 ~sh ~sparc x86 ~amd64-fbsd ~sparc-fbsd ~x86-fbsd"
+KEYWORDS="alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~s390 ~sh ~sparc x86 ~amd64-fbsd ~sparc-fbsd ~x86-fbsd"
 IUSE="caps debug python seccomp"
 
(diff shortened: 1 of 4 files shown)


pothos (commented):
I think you can add new entries in `coreos-overlay` under `profiles/coreos/base/package.accept_keywords` to achieve the same without modifying upstream ebuild files.

pothos:
These files are not expected to be changed and the next person that syncs them from Gentoo will run into the same problem again. Therefore, I think it makes sense to continue to add unstable packages to https://github.com/kinvolk/coreos-overlay/blob/main/profiles/coreos/base/package.accept_keywords

t-lo:
> These files are not expected to be changed and the next person that syncs them from Gentoo will run into the same problem again. 

But that's what I'm saying - they will not. All packages I've unmasked have newer ebuild files in Gentoo upstream which are also unmasked for ARM64. None of the files I touched are available upstream anymore.

So people syncing from Gentoo will use the new upstream ebuild unmasks, and everything will work.

pothos:
Yes, but when the new upstream ebuild file marks this package as unstable we would need to modify it again. Maybe it's ok since an error will be thrown and the person can add the entry to `accept_keywords` then.

t-lo:
I feel using `accept_keywords` is a bit of a hack and I hope we can reduce the usage eventually as we get closer to gentoo upstream, to a point where we don't need to override upstream masks at all but use their stable packages instead.

pothos (approved):

sayanchowdhury (approved):
I've usually follow the git commit logs to build my context. Would be great if you add a line about why we did not update `accept_keywords`. Other than looks good :shipit:

t-lo:
Amended the commit message to elaborate on why packages are unmasked here.

PR #181, opened 2021-06-28
sys-apps/gentoo-functions: use tarball from github

t-lo:
This one-line change pulls the gentoo-functions source tarball from
github instead from gentoo's own gitweb, to stabilise the build process.
We assume github to have higher availability than gentoo gitweb.

Signed-off-by: Thilo Fromm <thilo@kinvolk.io>


pothos (approved):
This URL won't be used as long as https://mirror.release.flatcar-linux.net/portage-stable/distfiles/ has the file.

Didn't verify the checksum but ok to merge.

t-lo:
I checked - well, emerge checked, I emerged `gentoo-functions` - works fine.

PR #183, opened 2021-07-01
app-misc/pax-utils: update to upstream 1.3.1

t-lo:
This change syncs pax-utils to 1.3.1, bringin it on par with Gentoo upstream. A minor change is incuded in the ebuild (commented on in the file) to work around a build issue with our (outdated) python-single-r1.eclass.

The workaround may be removed after we updated to a python version supported by upstream, and updated our eclasses respectively.

The update fixes an issue with `scanelf` and `glibc-2.33`:
```
/usr/lib/portage/python3.6/estrip: line 393: 1628751 Bad system call         (core dumped) scanelf -yqRBF '#k%F' -k '.symtab' "$@"
```

CI :large_blue_circle: : http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2946/cldsv/


dongsupark:
Good catch!
Tested and it indeed fixes the glibc 2.33 issue.
Looks like its upstream bug report is https://bugs.gentoo.org/709794.

pothos (approved):

PR #186, opened 2021-07-07
net-libs/libslirp: Import from Gentoo

jepio:
# net-libs/libslirp: Import from Gentoo

This is a dependency of qemu[slirp], which allows running flatcar images in qemu (and by extension: kola) from within the chroot.

## Testing done

`emerge libslirp`


sayanchowdhury:
Jenkins is running http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/

tormath1:
thanks ! I guess we also need to enable `slirp` USE flag in [sdk/package.use](https://github.com/kinvolk/coreos-overlay/blob/main/profiles/coreos/targets/sdk/package.use) then :)

jepio:
Yes, I have that as the next PR, but my process for submissions is not so streamlined yet :)
________________________________

dongsupark (approved):
Looks good, thanks!

Please note though, `net-libs/libslirp` is not included in `coreos-base/hard-host-depends`, so the SDK build does not automatically pull in libslirp by default.
As a result, the CI test did not build libslirp at all, I think.
Anyway I have just tested manually, and there was no build issue, so that's ok.

Later we would probably need to enable `slirp` USE flag in package.use.
Then we would not need to add libslirp explicitly to hard-host-depends.

PR #206, opened 2021-09-02
sys-libs/talloc: Move to overlay

krnowak:
We need to update sys-libs/ldb to drop the dependency on python2, but the updated package requires a newer version of sys-libs/talloc, which depends on a too new version of python3. Thus move the package to overlay.

To be merged together with https://github.com/kinvolk/coreos-overlay/pull/1248.

CI: http://localhost:9091/job/os/job/manifest/3479/cldsv/


krnowak:
CI passed.

pothos (approved):

PR #212, opened 2021-09-09
Initial steps for python3 transition

krnowak:
This PR makes some small changes in `python-utils-r1.eclass`: to support newer versions of python and to support invoking python_gen_cond_dep with only one parameter. These are more or less backports from the recent version of the eclass from gentoo, but I didn't just copy the eclass from there into portage-stable, because it drops support for old EAPIs that we still use.

We also take some packages back from overlay.

Needs to be merged together with https://github.com/kinvolk/coreos-overlay/pull/1269.

CI: http://localhost:9091/job/os/job/manifest/3546/


krnowak:
CI passed (had to rerun os/kola/qemu job though).

dongsupark (approved):

PR #214, opened 2021-09-14
Python transition: update python packages

krnowak:
~This pulls in #212, so it's blocked on it.~

Here mostly package updates happen, that drop compatibility with python2. Some packages are dropped, because they became unused. Some new packages come in as dependencies.

~I also updated virtual/pkgconfig that replaces dev-util/pkgconfig with dev-util/pkgconf. This happened as a result of updating dev-libs/libxslt.~

Another change here is about adding a hack to handle PYTHON_MULTI_USEDEP pseudovariable that is already deprecated in gentoo, but our old eclass files didn't handle it, so building some of the python packages failed.

Please see the list of commits to see what packages got added/droppped/updated.

One package that is modified more than usual is dev-python/boto, which was failing in CI, but not locally. Depending on how update to a newer version of python and eclasses will go, I'll either move boto to overlay or I'll revert the changes.

Comes with https://github.com/flatcar-linux/coreos-overlay/pull/1277 and https://github.com/flatcar-linux/scripts/pull/152.

CI: http://localhost:9091/job/os/job/manifest/4229/cldsv/


dev-libs/libxslt/libxslt-1.1.34-r1.ebuild
```diff
@@ -0,0 +1,67 @@
+# Copyright 1999-2021 Gentoo Authors
```
dongsupark, line 1: A minor note:
When you update or rebase this PR next time, can you please bump libxslt to 1.1.34-r2?
We should address CVE-2021-30560, which was fixed in the r2.
See also https://bugs.gentoo.org/820722 .

krnowak: Sure, will do.

krnowak: Done. Updated libxml2 at the same time, as comments in both say that they should be updated together.

krnowak:
CI passed.

krnowak:
Please hold on with reviews on this one - after talking with @jepio I decided to also put a python version update into it. That way I could drop all the commits that add support for python 3.6.

krnowak:
I needed to do a small update of `profiles/updates` to handle fully the rename of `dev-util/dwarves` into `dev-util/pahole` - without it, `update_chroot` didn't work. I still need to go through the changelog, because I think I got some versions wrong.

jepio (approved):
LGTM, thanks for updating so many packages.

PR #216, opened 2021-09-21
add dependencies for polkit 0.119 (with duktape)

wrl:
Adds:

 - acct-group/polkitd
 - acct-user/polkitd
 - dev-lang/duktape 2.6.0

Future PR will remove Spidermonkey (if it's not needed by any other packages).

CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3677/cldsv/


krnowak (commented):
One thing actually: for portage-stable, we usually copy the package verbatim, so `dev-lang/duktape` could also have both `duktape-2.4.0.ebuild` and `duktape-2.5.0-r1.ebuild` files.

krnowak:
> @krnowak Noted for the future. Do we care about that for this PR right now?

Yeah, for packages in portage-stable usually do something like:

```bash
cd dev-lang
git rm -rf duktape # just to remove an old version if there was any
cp -a "${GENTOO_REPO}/dev-lang/duktape" .
git add duktape
git commit -m 'dev-lang/duktape: Sync with gentoo'
```

On the other hand, in coreos-overlay, I prefer to keep just one ebuild, so it's clear which one is modified by us.

> I haven't tested with either of the older versions of duktape.

No need to test the older versions of duktape - the latest version is already declared stable (keywords for amd64 and arm64 do not have `~`), so portage will pick it up. Having older versions of duktape is simply a desire to have vanilla copies of gentoo packages in portage-stable.

> This will be accompanied by an updated PR for polkit itself, but I'm struggling with one of the downstream patches to the ebuild presently.

Alright. Please let me know if you need help there.

wrl:
@krnowak Spoke with Thilo about this and he said that including all upstream ebuilds isn't a hard and fast rule.

PR #218, opened 2021-09-24
Update kernel-2 eclass to support newer eclasses.

krnowak:
We also patch it to support python 3.6.

Needs to be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1303.

CI: http://localhost:9091/job/os/job/manifest/3661/cldsv/


krnowak:
CI passed.

krnowak:
I'm updating linux-headers package, so it uses EAPI 7 too. That way kernel-2 eclass does not barf at the old EAPI that our old linux-headers package used. The update should be fine I think, because we will install headers 5.10 which matches the kernel version in alpha. All the newer ones from 5.11 to 5.14 are keyworded as unstable.

PR #219, opened 2021-09-27
*: Drop dev-python/openssl and its dependencies

krnowak:
Our google-cloud-sdk dropped its requirement on pyopenssl, and it was the only package pulling it, so drop the pyopenssl package and some of the dependencies pulled only by pyopenssl.

CI: http://localhost:9091/job/os/job/manifest/3678/cldsv/


krnowak:
CI passed.

tormath1 (approved):
Thansk!

PR #227, opened 2021-10-19
Sync pkg-config implementation with gentoo

krnowak:
dev-util/pkgconfig is replaced by dev-util/pkgconf.

I needed to update some eclasses (systemd.eclass, udev.eclass), but at the same time I had to patch them to keep supporting the old EAPIs that we still use in portage-stable. I needed to patch bash_completion-r1.eclass to use the same fix that other eclasses have - this is something I think could be upstreamed to gentoo later.

This came up when updating python packages (I needed to update libxslt, which pulled in a newer version of virtual/pkgconfig, which obsoleted pkgconfig in favor of pkgconf, which caused some build issues during cross-compilation).

To be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1357.

CI: http://localhost:9091/job/os/job/manifest/4140/cldsv/


krnowak:
CI more or less passed (I'll keep rerunning qemu-uefi).

krnowak:
qemu-uefi tests passed too.

dongsupark:
I am still reading the PRs.
It looks slightly confusing, because we reverted multiple places to support old EAPIs.
I think we should create an issue to specify exactly which commits need to be reverted, and when it should be done.
Otherwise it looks quite easy to be forgotten.

krnowak:
> I am still reading the PRs. 

Thanks for doing it.

> It looks slightly confusing, because we reverted multiple places to support old EAPIs. I think we should create an issue to specify exactly which commits need to be reverted, and when it should be done. Otherwise it looks quite easy to be forgotten.

The only changes made was to not bail out on old EAPIs. So when we get rid of the old EAPIs (4 and 5 are left), we can simply sync the eclasses again. Not sure if I'll forget about it - my attempts at getting rid of old EAPIs is also motivated by being able to update the eclasses.

Oh, and that reminds me - there is an issue to file actually. It's about transitioning from `systemd_dotmpfilesd` from `systemd.eclass` to `dotmpfiles` from `tmpfiles.eclass`. The former got deprecated (and even recently removed) in favor of the latter. This will mostly involve our ebuild in coreos-overlay, as we use tmpfiles in a lot of places.

krnowak:
Filed an issue about tmpfiles - https://github.com/flatcar-linux/Flatcar/issues/546

eclass/systemd.eclass
```diff
@@ -405,6 +461,29 @@ systemd_is_booted() {
...
+#
+# Invokes systemd-tmpfiles --create with given arguments.
+# Does nothing if ROOT != / or systemd-tmpfiles is not in PATH.
+# This function should be called from pkg_postinst.
+#
+# Generally, this function should be called with the names of any tmpfiles
+# fragments which have been installed, either by the build system or by a
+# previous call to systemd_dotmpfilesd. This ensures that any tmpfiles are
+# created without the need to reboot the system.
+systemd_tmpfiles_create() {
```
dongsupark: Why should we bring back the function `systemd_tmpfiles_create()`?
Looks like the function is not used anywhere, right?

krnowak: Hm, I just brought back everything related to tmpfiles. But yeah, this one is actually unused, I'll drop it.

krnowak: Removed.

dongsupark (approved):
Thanks.

Before merging it, how about squashing the commit https://github.com/flatcar-linux/portage-stable/pull/227/commits/5608c918acba8a27192964cf3597e6482b78eaa0 to the corresponding one https://github.com/flatcar-linux/portage-stable/pull/227/commits/682576c520df6649333324206460525b417395cf ?

krnowak:
> Before merging it, how about squashing the commit [5608c91](https://github.com/flatcar-linux/portage-stable/commit/5608c918acba8a27192964cf3597e6482b78eaa0) to the corresponding one [682576c](https://github.com/flatcar-linux/portage-stable/commit/682576c520df6649333324206460525b417395cf) ?

Sure, will do. Thanks for the review.

PR #231, opened 2021-10-24
Remove app-admin/logrotate from portage-stable

aniruddha2000:
# Remove app-admin/logrotate from portage-stable

As the new logrotate files from upstream gentoo is not suitable for the Flatcar. We are moving this package to coreos-overlay. For that we are removing the package directory from this repository.

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4098/cldsv


krnowak (approved):
Looks good, but I think it's blocked on my question in https://github.com/flatcar-linux/init/pull/52.

PR #234, opened 2021-10-29
script/bootstrap: update openssl before stage3

jepio:
# script/bootstrap: update openssl before stage3

Depclean failures in stage3 have been tracked down to stage1 containing openssl-1.1, and stage3 packages being built against that (dependencies are soft enough that they are not ordered against openssl-3 update in stage3). Add a workaround: explicitly build openssl-3 in stage2, right before stage3.

## How to use

`./boostrap_sdk`.

## Testing done

CI will run.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -278,6 +278,8 @@ for atom in portage.settings.packages:
 [[ -z ${myNCURSES}    ]] && myNCURSES="sys-libs/ncurses"
 # Flatcar: install curl with BOOTSTRAP_USE=ssl to fetch from https URLs
 [[ -z ${myCURL}       ]] && myCURL="net-misc/curl"
+# Flatcar: upgrade to openssl-3 before system rebuild in stage3
+[[ -z ${myOPENSSL}    ]] && myOPENSSL="dev-libs/openssl"
 
 # Do we really want gettext/nls?
 [[ ${USE_NLS} != 1 ]] && myGETTEXT=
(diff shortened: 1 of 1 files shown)


dongsupark:
Looks good if CI passed.

But can you please add an `einfo` message for openssl [here](https://github.com/flatcar-linux/portage-stable/blob/e20d2e4d478ef76de7dba0a04e39414914586d19/scripts/bootstrap.sh#L301)?

jepio:
I rebased to resolve merge conflicts with the curl addition and added the einfo call. I wanted to hold off on merging to observe some more, but this still happens occasionally in testing so merging it now.

dongsupark:
Thanks for merging it.

Site note: When we have some time, we should fix the same issues in upstream Gentoo.
Including this PR and https://github.com/flatcar-linux/portage-stable/pull/240/commits/876ddceb688a4bb96b59a131ccecdfbe347dc406 .

tormath1:
@dongsupark great idea. Is there an existing bug report to ease the reviewing process ?

dongsupark:
Not sure. But there is a similar issue https://bugs.gentoo.org/817914 as the curl ssl issue.

jepio:
The openssl issue is self-inflicted because our stage1 does not use the same ebuilds as stage2-3, so that one does not need to go upstream.

It would make sense to push the curl fix upstream though. Gentoo is probaby not hitting it because all sources needed for bootstrap are available on the mirror?

dongsupark:
> Gentoo is probaby not hitting it because all sources needed for bootstrap are available on the mirror?

Right.
While flatcar-linux/baselayout fetches from GIthub, the Gentoo's baselayout does not touch Github. That's why the issue has been hidden in Gentoo.
https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-apps/baselayout/baselayout-9999.ebuild#n14

PR #235, opened 2021-11-01
dev-lang/spidermonkey: delete unnecessary ebuilds

dongsupark:
https://github.com/flatcar-linux/coreos-overlay/pull/1263 made polkit depend on duktape instead of spidermonkey.
So we can delete spidermonkey completely.

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1384 .

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4074/cldsv


krnowak:
Could you also drop the unnecessary eclasses too? I think you could drop mozconfig, mozcoreconf, mozlinguas and nsplugins.

dongsupark:
Thanks, deleted unnecessary eclass files as well.

krnowak (approved):

PR #245, opened 2021-11-10
app-editors/nano: update to 5.8-r2

dongsupark:
Now that ncurses 6.2_p20210619 does not have USE flag `unicode` at all, `app-editors/nano` pulling in ncurses with `USE=unicode` results in build failures.

We should have `unicode(+)` to make the flag optional like upstream Gentoo does.

Simply sync with Gentoo for nano 5.8-r2 to fix that issue.

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1410.

## Testing done

CI passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4128/cldsv
(with specific SEED_SDK_{OVERLAY,PORTAGE}_GITREF values)


sayanchowdhury (approved):

dongsupark:
cherry-picked to flatcar-3046

PR #246, opened 2021-11-10
Update packages to get rid of EAPI 4

krnowak:
Please see the commits to see what packages got updated. I also dropped two eclasses that weren't used anywhere.

Should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1413.

CI: http://localhost:9091/job/os/job/manifest/4408/cldsv/


krnowak:
CI passed.

sys-apps/pv/pv-1.6.19-r1.ebuild
```diff
@@ -11,7 +11,7 @@ SRC_URI="https://www.ivarch.com/programs/sources/${P}.tar.bz2"
```
dongsupark, line 11: I think we do not need to downgrade pv to 1.6.19, as we already have `sys-apps/portage` 3.0.28, right?

dongsupark (commented):
Mostly looks good.

We would still need changelogs for the package updates.

sys-apps/pv/pv-1.6.19-r1.ebuild
krnowak: Oh yeah, I can update that now. Thanks for spotting.

krnowak:
> Mostly looks good.
> 
> We would still need changelogs for the package updates.

There is already a changelog -  https://github.com/flatcar-linux/portage-stable/pull/246/files#diff-6a01b3f753a3e3d2020766b79d5e301b0e2d9a2c159c29c2d58bf2e82e10ca64 (if the link doesn't work because of updated PR, then searching for `changelog/updates` should work).

sys-apps/pv/pv-1.6.19-r1.ebuild
krnowak: I actually had two commits updating `sys-apps/pv`. First was updating it to recent version, the second was downgrading it a bit. Dropped the latter.

dongsupark:
Thanks, looks good to me.
I would still let @sayanchowdhury and @tormath1 have another look, as it is a huge change.

krnowak:
> Thanks, looks good to me. I would still let @sayanchowdhury and @tormath1 have another look, as it is a huge change.

Sure. Thanks for the review.

sys-apps/debianutils/debianutils-5.4.ebuild
```diff
@@ -0,0 +1,45 @@
...
+HOMEPAGE="https://packages.qa.debian.org/d/debianutils.html"
+SRC_URI="mirror://debian/pool/main/d/${PN}/${PN}_${PV}.orig.tar.xz"
+
+LICENSE="BSD GPL-2 SMAIL"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~x86-linux"
+IUSE="+installkernel static"
+
+PDEPEND="
+	installkernel? (
```
tormath1, line 18: I just have a doubt on this dependency:
1. It's a `PDEPEND` one
2. I'm not sure we need `installkernel` script to install kernel, it's already handled by the kernel ebuild itself I guess ?

krnowak: > I just have a doubt on this dependency:

I had my doubt about why we even have these debianutils in SDK. Personally I'd just drop the package, but maybe someone is actually using something from it. [The debianutils package website](https://packages.debian.org/sid/debianutils) tells me that it contains the following tools: add-shell installkernel ischroot remove-shell run-parts savelog update-shells which. Of those, probably `which` is "interesting", but it could be easily replaced with bash's builtin `command -v`.

> 
>     1. It's a `PDEPEND` one

And what's wrong with it? There must be something I'm not getting.

>     2. I'm not sure we need `installkernel` script to install kernel, it's already handled by the kernel ebuild itself I guess ?

We always had an (most likely broken anyway) `installkernel` utility in the SDK, but with the update, it seems to be split out into a separate package. So I was basically trying to preserve what we had so far in SDK.

I guess I could either force disable the installkernel USE flag, or just drop this package altogether. What do you think?

dev-cpp/glog/glog-0.5.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc64 ~s390 ~x86 ~amd64-linux ~x86-linux"
+IUSE="gflags +libunwind llvm-libunwind test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="gflags? ( dev-cpp/gflags:0=[${MULTILIB_USEDEP}] )
+	libunwind? (
```
tormath1, line 30: Maybe we could deactivate `libunwind` support - `glog` is only used by `update-engine` I think ? 

Reading `glog` documentation[^1]:
> For that reason, if you use a 64-bit system and you need InstallFailureSignalHandler(), we strongly recommend you install libunwind before trying to configure or install google glog

Quick search on [`update-engine`](http://github.com/flatcar-linux/update-engine) repository does not show any sign of `InstallFailureSignalHandler`.

[^1]: https://github.com/google/glog#installation-notes-for-64-bit-linux-systems

krnowak: I think we are pulling in glog 0.4, which pulls libunwind unconditionally. 0.5 is still marked as unstable.

sys-apps/debianutils/debianutils-5.4.ebuild
tormath1: > installkernel utility in the SDK, but with the update, it seems to be split out into a separate package.

Oh ok, then it might be safe to keep it - we can try to remove it in a following PR to not block this one furthermore. :) 

> And what's wrong with it? There must be something I'm not getting.

To me, `PDEPEND` means "this dependency does not impact the build / the run of this program" so I doublecheck the utility of a dependency with it's defined as `PDEPEND`

krnowak: > > installkernel utility in the SDK, but with the update, it seems to be split out into a separate package.
> 
> Oh ok, then it might be safe to keep it - we can try to remove it in a following PR to not block this one furthermore. :)
> 

I'll try to drop the debianutils package later - I've seen a use of `which` in our build scripts, so maybe those will need a migration. In general I'd say that our SDK needs a check at what it's pulling, because it seems to be that we carry there some unnecessary stuff.

> > And what's wrong with it? There must be something I'm not getting.
> 
> To me, `PDEPEND` means "this dependency does not impact the build / the run of this program" so I doublecheck the utility of a dependency with it's defined as `PDEPEND`

Ah, thanks! Will keep that in mind.

changelog/updates/2021-12-17-eapi4-update.md
```diff
@@ -0,0 +1,8 @@
+- ipset ([7.11](https://ipset.netfilter.org/changelog.html))
```
tormath1, line 1: I guess missing packages in the changelog concern the SDK and not the end user ?

tormath1 (approved):
One last question - otherwise it looks really good to me. :rocket:

changelog/updates/2021-12-17-eapi4-update.md
krnowak: Yes, the unmentioned packages are not user facing.

PR #250, opened 2021-11-19
changelog: Add placeholder directory to add the changelogs

sayanchowdhury:


changelog/README.md
```diff
@@ -0,0 +1,15 @@
...
+- security: List of PRs fixing security issues
+- updates: Package Updates
+
+## How to add the file
+
+Based on the category the PR falls into create a new file in the respective directory
+with the filename format `DD-MM-YYYY-<few-words-about-the-change>.md`
+
+The contents of the file should describe the changes in an ellaborative manner, except for `Updates`.
+As `Updates` refer to the pacakage updates, the description of the file should be of the
```
tormath1, line 14:
```suggestion
As `Updates` refer to the package updates, the description of the file should be of the
```

changelog/README.md
```diff
@@ -0,0 +1,15 @@
...
+The changes are essentially divided into 4 categories
+- changes: List of PRs brings Changes and/or Enhancements
+- bugfixes: List of PRs fixing existing issues
+- security: List of PRs fixing security issues
+- updates: Package Updates
+
+## How to add the file
+
+Based on the category the PR falls into create a new file in the respective directory
+with the filename format `DD-MM-YYYY-<few-words-about-the-change>.md`
```
tormath1, line 11: We can suggest a help to generate the date format:

```suggestion
with the filename format `DD-MM-YYYY-<few-words-about-the-change>.md (can be generated via: `$(date '+%d-%m-%Y')-<few-words-about-the-change>.md``
```

changelog/README.md
```diff
@@ -0,0 +1,15 @@
+The `changelog` directory contains the description of the changes introduced into the repository.
+The changes are essentially divided into 4 categories
```
tormath1, line 2:
```suggestion
The changes are essentially divided into 4 categories:
```

changelog/README.md
```diff
@@ -0,0 +1,15 @@
...
+- changes: List of PRs brings Changes and/or Enhancements
+- bugfixes: List of PRs fixing existing issues
+- security: List of PRs fixing security issues
+- updates: Package Updates
```
tormath1, line 6:
```suggestion
- changes: PRs bringing Changes and/or Enhancements
- bugfixes: PRs fixing existing issues
- security: PRs fixing security issues
- updates: PRs updating packages
```

changelog/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## How to add the file
+
+Based on the category the PR falls into create a new file in the respective directory
+with the filename format `DD-MM-YYYY-<few-words-about-the-change>.md`
+
+The contents of the file should describe the changes in an ellaborative manner, except for `Updates`.
```
tormath1, line 13: We could maybe add an entry example:

```md
$ cat `changelog/bugfixes/(date '+%d-%m-%Y')-selinux-torcx.md
add SELinux rule to allow torcx to correctly start in SELinux enforced mode ([#1426](https://github.com/flatcar-linux/coreos-overlay/pull/1426))
```
Since we give an `Update` example right after, that would make sense. :)

tormath1:
It's just general suggestion :arrow_up: Glad to have this into the main Flatcar repository :D

pothos:
Looks good, I think we need the top level `CHANGELOG.md` file only in the flatcar-scripts repo which will get the other repos as submodules and could be a good place to generate the final file (currently the generator would need to get the path to portage-stable/coreos-overlay). One thing to note is that the final file is different from the website content - it has all changes per release while on the website we list the releases changes per channel with an additional entry about the changes compared to a previous channel releases when a transition happens. We can even discuss to skip generating the final file but I think it has a value to see the changes that happened incrementally up to this release while on the website we can't see this well due to the channel split which forces to switch the tabs.

changelog/README.md
pothos: Do you also have an example how to use flog to generate an update entry with a link to the changelog or an CVE list entry? Or is this not ready yet for this use case?

krnowak:
```suggestion
The contents of the file should describe the changes in an elaborative manner, except for `Updates`.
```

changelog/README.md
```diff
@@ -0,0 +1,15 @@
...
+As `Updates` refer to the pacakage updates, the description of the file should be of the
+following format: `Package Name [Version](link to changelog)`
```
krnowak, line 15: I would like to see an example of it. I get the `Package Name` part. It would be something like `dev-util/perf`. `Version` would be version taken from ebuild filename, right (so `5.15-r1`)? I don't get the `link to changelog` part - what changelog?

pothos: It means the release notes of the updated package

krnowak: Could be tricky for some projects. What would use for `dev-util/perf`? https://kernelnewbies.org/Linux_5.15#Tracing.2C_perf_and_BPF for example?

t-lo: Package name is the upstream name, no need to prefix the group. Also, version is the upstream release version w/o recipe suffix. Changelog is a link to the upstream release changelog (if available).
A good example would be previously updated item from an existing changelog, e.g. Linux in the latest stable release:

```
# Updates
- Linux ([5.10.77](https://lwn.net/Articles/874852/))
```

krnowak: Cool, thanks for the example. I think it needs to be clarified in the README.

changelog/README.md
sayanchowdhury: I did type the correct spelling of elaborative, not sure why vim suggested me double-l 😖

sayanchowdhury: The intention of `flog` is different, it would be a central place to all the project changelog. My idea is to have a script to do the release, something like [blurb](https://pypi.org/project/blurb/) but for us, it would be just a small script.

changelog/README.md
```diff
@@ -0,0 +1,15 @@
...
+- bugfixes: PRs fixing existing issues
+- security: PRs fixing security issues
+- updates: PRs updating packages
+
+## How to add the file
+
+Based on the category the PR falls into create a new file in the respective directory
+with the filename format `DD-MM-YYYY-<few-words-about-the-change>.md (can be generated via: `$(date '+%d-%m-%Y')-<few-words-about-the-change>.md``
+
+The contents of the file should describe the changes in an elaborative manner, except for `Updates`.
```
pothos, line 13:
```suggestion
The contents of the file should describe the changes in an elaborative manner (use the past tense for the change/bugfix description to avoid confusion with the imperative voice for actions the user should do as a result).
Security fixes of upstream packages and package updates can be kept short in most cases and follow a standard format.
```

pothos (approved):
I think this it good to go and can be changed based on experience from starting to use it :)

PR #251, opened 2021-11-25
Sync some packages with linux kernel 5.15

krnowak:
This pulls in newer versions of sys-apps/iproute2, dev-util/perf and sys-kernel/linux-headers. They are not marked as stable so we add the relevant keywords in the coreos-overlay PR.

To be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1443.

CI: http://localhost:9091/job/os/job/manifest/4263/cldsv/


krnowak:
CI passed.

jepio (approved):
Looks OK, the only question I had was around dev-util/perf and whether we should move it to coreos-overlay over the one line change?

krnowak:
@jepio : I actually remember doing it in the beginning, but it seems that the agreement is that PYTHON_COMPAT changes can stay in portage-stable. Hopefully soon those won't be necessary anyway.

jepio:
OK - im fine with that interpretation

krnowak:
Changelogs added, merging.

PR #253, opened 2021-11-26
changelog: use date format of YYYY-MM-DD

dongsupark:
If we use date format of `DD-MM-YYYY` in changelog file names, the files will not sorted by date. e.g. `01-12-2021` will come before `25-11-2021`.

Use date format of `YYYY-MM-DD` to make the files sorted by date.


pothos (approved):
Thanks, you can push this change to the other repos, too

t-lo (approved):
LGTM - thank you for the quick turnaround, Dongsu!

PR #256, opened 2021-12-02
sys-firmware: update edk2-ovmf, ipxe, seabios, sgabios

dongsupark:
Update the following packages needed by qemu 6.1.0.

* edk2-ovmf 202105-r1
* ipxe 1.21.1
* seabios 1.14.0
* sgabios 0.1_pre10

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1469 .

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4308/cldsv


sys-firmware/edk2-ovmf/edk2-ovmf-202105-r1.ebuild
```diff
@@ -1,49 +1,63 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 PYTHON_REQ_USE="sqlite"
-PYTHON_COMPAT=( python{3_6,3_7} )
+PYTHON_COMPAT=( python3_{8,9,10} )
```
krnowak, line 7: I think you will need to preserve python 3.6 still, though.

sys-firmware/seabios/seabios-1.14.0-r2.ebuild
```diff
@@ -0,0 +1,183 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
+
+PYTHON_COMPAT=( python3_{7..9} )
```
krnowak, line 6: I think you will need to preserve python 3.6 compatibility here, though.

sys-firmware/edk2-ovmf/edk2-ovmf-202105-r1.ebuild
dongsupark: Yeah, that's interesting.
I was wondering why the build passed, even if I did not touch `PYTHON_COMPAT` in both edk2-ovmf and seabios.

Both packages inherit `python-any-r1.eclass`, a relatively minimal eclass.
There `PYTHON_REQUIRED_USE` is not so strictly checked, and the builds passed.

On the other hand, for example, when an ebuild inherits `distutils-r1.eclass`, then the `PYTHON_REQUIRED_USE` is a [hard requirement](https://github.com/flatcar-linux/portage-stable/blob/16a083eb50202ccbca7f7368ce9fc1f4faab5aa8/eclass/distutils-r1.eclass#L154).
In that case, without modifying PYTHON_COMPAT, it fails like
`The following REQUIRED_USE flag constraints are unsatisfied: any-of ( python_targets_pypy3 python_targets_python3_7 python_targets_python3_8 python_targets_python3_9 )`.

Does it make sense?
Is it ok just leave the ebuilds like that, without moving to coreos-overlay?

krnowak: I'd say that if the build passes then it's fine to keep it without modifications. It will become a non-problem soon, when we land my python update work.

krnowak (approved):

dongsupark:
CI passed

PR #255, opened 2021-12-01
app-editors/vim: update to 8.2.3582

dongsupark:
Update `app-editors/vim` and `app-editors/vim-core` to 8.2.3582.
Import `dev-libs/libsodium` needed by vim[-minimal], only for SDK.

It is mainly to address the following security issues:
* [CVE-2021-3872](https://nvd.nist.gov/vuln/detail/CVE-2021-3872)
* [CVE-2021-3875](https://nvd.nist.gov/vuln/detail/CVE-2021-3875)
* [CVE-2021-3903](https://nvd.nist.gov/vuln/detail/CVE-2021-3903)
* [CVE-2021-3927](https://nvd.nist.gov/vuln/detail/CVE-2021-3927)
* [CVE-2021-3928](https://nvd.nist.gov/vuln/detail/CVE-2021-3928)
* [CVE-2021-3968](https://nvd.nist.gov/vuln/detail/CVE-2021-3968)
* [CVE-2021-3973](https://nvd.nist.gov/vuln/detail/CVE-2021-3973)
* [CVE-2021-3974](https://nvd.nist.gov/vuln/detail/CVE-2021-3974)

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1455.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4278/cldsv

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)

PR #260, opened 2021-12-09
app-editors/vim: sync with gentoo to fix missing defaults.vim

dongsupark:
vim >= 8.2.3428 prints out a warning on missing file, when it is built with USE="minimal" and `app-editors/vim-core` is not installed.

```
$ vim
E1187: Failed to source defaults.vim
Press ENTER or type command to continue
```

To remove warning, explicitly install /usr/share/vim/vim82/defaults.vim.

This issue was fixed via https://github.com/gentoo/gentoo/pull/23115.
That fix was already included in https://github.com/flatcar-linux/portage-stable/pull/255, but only for main.
That's why we need to fix it also in maintenance branches.

Also add corresponding fixes for vim-core and libsodium.

Gentoo ref: https://github.com/gentoo/gentoo/commit/fc72192754c8a1ff9014fd739a2d07056b25ed1e, https://github.com/gentoo/gentoo/commit/74eaf6b7cbab24337eae5c07d42871aeb21a330e .

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1493.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4355/cldsv


dongsupark:
CI passed

krnowak:
One question: why do we need libsodium in portage-stable if you disabled the crypt use flag?

dongsupark:
Because the `crypt` USE flag is still enabled in the SDK profile.
SDK: USE=`crypt -minimal`. vim, vim-core, libsodium installed
amd64 & arm64: USE=`-crypt minimal`. only vim installed. vim-core and libsodium are not installed.

In theory we could simply remove libsodium from SDK as well.
But unlike the board packages, SDK has always had a full-blown version of vim, so I thought we would rather keep the crypt USE flag and libsodium.

krnowak:
Thanks. Let's keep it like that then.

krnowak (approved):
Looks good.

dongsupark:
cherry-picked to flatcar-3033.

PR #262, opened 2021-12-21
eclass: Sync some with gentoo to support EAPI 8, but keep supporting EAPI 4 too

krnowak:
To be merged with https://github.com/flatcar-linux/coreos-overlay/pull/1519.

Needed to update `dev-util/cmake` too, because the updated cmake eclass bumped its required version. Also updated `sys-block/parted` to fix build issues with mismatching gettext infra.

CI: http://localhost:9091/job/os/job/manifest/4540/cldsv/


krnowak:
Needed to update sys-block/parted as it is old enough to break with some gettext stuff because of the eclass updates. This will need a changelog update, as parted is a part of the production image.

krnowak:
CI passed.

jepio (approved):
LGTM, hope we can get rid of EAPI 0/4 soon.

krnowak:
> LGTM, hope we can get rid of EAPI 0/4 soon.

Thanks for the review. Most of EAPI 4 is gone (one package left). EAPI 0 is still a thing, unfortunately.

PR #263, opened 2021-12-21
Drop a bunch of unused packages

krnowak:
Nothing changeloggable here. Just dropping some dead weight.

CI passed: http://localhost:9091/job/os/job/manifest/4439/cldsv/


krnowak:
CI passed.

jepio (approved):

PR #267, opened 2022-01-05
Migrate to xcrypt

krnowak:
Gentoo migrated to libxcrypt to replace the deprecated crypt functionality in glibc. To do the same, we need to update profiles, virtual/libcrypt and add sys-libs/libxcrypt. The `compat` USE flag seems to be enabled by default, so it will install both libcrypt.so.1 (which used to be provided by glibc) and libcrypt.so.2 (which is the new stuff).

I'm not sure we need to keep compat stuff, but this probably can be revisited later.

Needs to be merged with https://github.com/flatcar-linux/coreos-overlay/pull/1543.

CI: http://localhost:9091/job/os/job/manifest/4524/cldsv/


krnowak:
Closing, explanation in coreos overlay PR linked above.

PR #269, opened 2022-01-11
eclass: Drop unused eclasses

krnowak:
Just dropping more unused stuff. Less stuff to update later.

CI: http://localhost:9091/job/os/job/manifest/4565/cldsv/


krnowak:
CI passed.

tormath1 (approved):
LGTM , bye `webapp` eclass ! Just curious, do you know from where this eclass comesfrom ? An initial sync with `::gentoo` ?

krnowak:
> LGTM , bye `webapp` eclass ! Just curious, do you know from where this eclass comesfrom ? An initial sync with `::gentoo` ?

Thanks for review.

My quick investigation tells me that at some point chromium folks pulled in all the eclasses from portage. There was even a package (`media-libs/portage/net-irc/cgiirc`, looks weird) that used it, but it was imported by mistake, so got dropped later. But there probably wasn't an eclass cleanup ever done, so webapp.eclass remained only because it's still a thing in gentoo.

PR #270, opened 2022-01-11
net-misc/iputils: Update the package

krnowak:
This is the last package with EAPI 4 to be updated.

Needs to be merged with https://github.com/flatcar-linux/coreos-overlay/pull/1556.

CI: http://localhost:9091/job/os/job/manifest/4576/cldsv/


krnowak:
CI passed.

pothos (approved):
nit: In general I think the changelog entry can be part of the other commit which makes cherry-picking easier

krnowak:
Thanks for the review.

> nit: In general I think the changelog entry can be part of the other commit which makes cherry-picking easier

Alright, I'll keep that in mind.

PR #272, opened 2022-01-17
Update some app-arch software

krnowak:
This is to remove EAPI 5 from portage-stable.

Not a lot of updates here, actually:
- cpio has version 2.13 but is marked as unstable, so not used - we still remain at version 2.12.
- lbzip2 is still at version 2.5
- pbzip2 is not user visible
- pigz same
- bzip2 is not updated too - it's still at version 1.0.8, we just move it from coreos-overlay back to portage-stable
- sharutils is not user visible

Thus, no changelog.

Needs to be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1573.

CI: http://localhost:9091/job/os/job/manifest/4611/cldsv/


krnowak:
CI passed.

tormath1 (approved):

PR #276, opened 2022-01-20
Add missing users and groups for systemd update, update them and their eclasses

krnowak:
Subject says it all.

Should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1579.

CI: http://localhost:9091/job/os/job/manifest/4843/cldsv/


krnowak:
CI passed.

pothos (approved):

PR #281, opened 2022-02-02
Drop duplicated packages

krnowak:
`app-crypt/sbsigntools` and `sys-process/tini` already are in overlay, so drop them.

CI: http://localhost:9091/job/os/job/manifest/4744/cldsv/


sayanchowdhury (approved):
:shipit:

PR #285, opened 2022-02-10
update meson

jepio:
# update meson

Update meson to a current version, which is needed when introducing clevis ebuilds.

## How to use

`./build_packages && ./build_image`

## Testing done

CI running: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4810/cldsv/


dongsupark:
Looks good if CI passed.

Can you please also delete rng-tools from https://github.com/flatcar-linux/coreos-overlay/blob/233d43aa144dfa508739d4b31644478919591cef/profiles/coreos/arm64/package.accept_keywords#L57 ?

tormath1:
Mm, polkit build is failing because of this I think: https://mesonbuild.com/Release-notes-for-0-60-0.html#unknown-options-are-now-always-fatal.
I think we could replace the useless `-Dwith-duktape=yes` by this option: https://gitlab.freedesktop.org/polkit/polkit/-/blob/master/meson_options.txt#L19. It was correctly built previously because the option seems to default to `duktape`.

jepio:
Thanks @tormath1 , can you submit the duktape fix as a PR? The remaining issue is that now the rngd service exits with error when hwrng is not available, so i'll adjust the `ExecStart` line to tolerate failure.

pothos:
Is rngd now starting by default? We had a discussion in in https://github.com/flatcar-linux/Flatcar/issues/207 about that. I think it won't hurt to try to run it and handle it gracefully. Another idea was to only start it based on seeing what `/dev/random` is backed by `/sys/class/misc/hw_random/rng_*` and whether rngd adds value or not but that is maybe a bit complex.

jepio:
Yes, we automatically enable rngd but only for the initramfs (https://github.com/dracutdevs/dracut/blob/master/modules.d/06rngd/module-setup.sh). `/sys/class/misc/hw_random/rng_*` is related to `/dev/hwrng` but these don't get fed into `/dev/random` automatically, that's what rngd does. rngd support multiple backends and on x86 it uses the rdrand instruction. The new version also can checks whether `/dev/hwrng` has an backend before using.

pothos:
Ok, should we enable it by default then in general instead of only the initramfs?

JAORMX:
uhm... isn't it the case that `rngd` is no longer needed? At least it doesn't seem to be shipping anymore in Fedora [[1]](https://pagure.io/fedora-workstation/issue/184) [[2]](https://pagure.io/fedora-comps/pull-request/533), and RHEL [[3]](https://bugzilla.redhat.com/show_bug.cgi?id=1888695)

pothos:
> uhm... isn't it the case that `rngd` is no longer needed? At least it doesn't seem to be shipping anymore in Fedora [[1]](https://pagure.io/fedora-workstation/issue/184) [[2]](https://pagure.io/fedora-comps/pull-request/533), and RHEL [[3]](https://bugzilla.redhat.com/show_bug.cgi?id=1888695)

With the new development I hope yes, it won't be needed anymore: https://lwn.net/Articles/884875/

jepio:
@JAORMX might be right - I will test removing rngd altogether.

jepio:
Dropped the rng-tools part because we removed it. Rerunning CI.

tormath1 (approved):
let's wait for the CI but it's a "go" for me.

dongsupark (approved):
Looks good if it has changelog as well.

PR #286, opened 2022-02-14
net-fs/cifs-utils: update to 6.13-r1

dongsupark:
Update net-fs/cifs-utils to 6.13-r1, mainly to address [CVE-2021-20208](https://nvd.nist.gov/vuln/detail/CVE-2021-20208).

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4853/cldsv


dongsupark:
Upstream PR was merged, CI passed with that change.

krnowak (approved):
Looks good, let's get this in - it's the last package that was dependent on `virtual/pam` that I wanted to remove today. Question though - is `-python` in `make.defaults` enough to disable `+python` in the ebuild file?

dongsupark:
> - is -python in make.defaults enough to disable +python in the ebuild file?

Yes, the python USE flag is disabled by default, so no need to explicitly do that again.

PR #287, opened 2022-02-16
Update packages from 2014

krnowak:
These are mostly virtual or SDK packages, so nothing visible for the end user. Dropped one obsolete virtual package and a bunch of unused eclasses.

There's one eclass from 2014 that wasn't yet handled - `autotools-multilib.eclass`. It's used by `xorg-2.eclass` and by `dev-libs/ding-libs`. It will eventually go away as we progress with package updates.

CI: http://localhost:9091/job/os/job/manifest/4924/cldsv/


krnowak:
CI passed.

dongsupark (approved):

PR #289, opened 2022-02-17
metadata: update GLSA metadata 2022-02-16

dongsupark:
Sync GLSA metadata with Gentoo, as of 2022-02-16.

This PR should be merged together with https://github.com/flatcar-linux/scripts/pull/229.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4867/cldsv

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) - not needed


krnowak:
Out of interest: what's the workflow with GLSA updates? I'll be moving those files to coreos-overlay at some point, so I'd like to know what else needs to be updated.

krnowak (approved):

dongsupark:
> Out of interest: what's the workflow with GLSA updates?

There is no strict rule about that.
Ideally the GLSA metadata should be synced with Gentoo as often as possible.
In the past, however, the sync resulted in extra work like having to add more entries in the allowlist. That's the reason why the sync did not happen that often.
At the moment, we do not need so many exceptional cases in the list, thanks to recent updates of python and perl especially.
So I think we can automate the GLSA metadata sync in the future.

krnowak:
So currently there is no automation or any scripts that would need updating after moving the GLSA metadata elsewhere?

dongsupark:
> So currently there is no automation or any scripts that would need updating after moving the GLSA metadata elsewhere?

Correct.
The [`glsa-check-$BOARD`](https://github.com/flatcar-linux/scripts/blob/221351927e10e0dbaacca9d530663b36030cd387/build_library/test_image_content.sh#L22) is the only caller.

krnowak:
> > So currently there is no automation or any scripts that would need updating after moving the GLSA metadata elsewhere?
> 
> Correct. The [`glsa-check-$BOARD`](https://github.com/flatcar-linux/scripts/blob/221351927e10e0dbaacca9d530663b36030cd387/build_library/test_image_content.sh#L22) is the only caller.

Cool, thanks for the pointer.

PR #292, opened 2022-02-17
dev-db/sqlite: move to `::coreos-overlay`

tormath1:
mainly to apply a patch to skip `sqlite3` cli installation.

See also: https://bugs.gentoo.org/833475

<hr>

no change log required


tormath1:
closing this, since we'll prevent the `sqlite3` binary installation directly from `INSTALL_MASK`.

PR #295, opened 2022-02-21
Update eclasses from 2013

krnowak:
No big changes though, other than dropping support for old EAPIs.

CI: http://localhost:9091/job/os/job/manifest/4893/cldsv/


krnowak:
CI passed.

jepio (approved):

PR #299, opened 2022-02-24
Update packages and eclasses from 2015

krnowak:
The only package from 2015 that's left out is sys-apps/rng-tools - it's being updated/dropped in #285. Of all the updated packages that are user visible there's only bridge-utils. I wasn't adding updated libraries to the changelog. Not sure if they should be, actually.

Also not updating the fcaps eclass - the app-cdr/cdrtools uses the old EAPI, so it needs to be updated first (will do in another PR, for 2016).

Needs to be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1674.

CI: http://localhost:9091/job/os/job/manifest/4960/cldsv/


krnowak:
CI passed.

pothos (approved):

PR #301, opened 2022-03-01
Update packages and eclasses from 2017

krnowak:
From user visible changes - it updates - whois, rpcbind, socat, dosfstools, thin-provisioning-tools, lsscsi and man-db.

~Blocked on #300.~

Needs to be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1681.

CI: http://localhost:9091/job/os/job/manifest/5080/cldsv/

PR #302, opened 2022-03-03
Update packages and eclasses from 2018

krnowak:
From user visible changes - it updates - pinentry, quota, xfsprogs, grep and less. Some libraries are also updated, but I didn't mention them in changelog. Not sure if I should.

Blocked on #301.

Should be merged together with COREOS_OVERLAY_PR.

CI: http://localhost:9091/job/os/job/manifest/5019/cldsv/


krnowak:
CI passed.

dongsupark (approved):

PR #311, opened 2022-03-22
sys-fs/cryptsetup: sync with ::gentoo

tormath1:
this pulls fips support.

Commit-Ref: https://github.com/gentoo/gentoo/commit/1746c2caf877e7b678a5028ecff4bfcfb8899ff1

- [ ] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5206/cldsv/


pothos:
Looks like we can also backport this to Stable for having it in the LTS, or?

tormath1:
> Looks like we can also backport this to Stable for having it in the LTS, or?

@pothos that sounds good to me. The only blocking point I see is the bump from 2.3.6 to 2.4.3 on Stable without going through alpha/beta.

pothos:
Oh, misread the versions. Ok, maybe we can backport this then.

pothos (approved):

tormath1:
@pothos I was about to open the PR to backport cryptsetup's fips mode to stable then I realized that we'll still miss OpenSSLv3 built with fips provider in stable anyway..

pothos:
I've checked and LTS didn't use OpenSSL with FIPS yet. I think just the cryptsetup flag is enough as portage-stable patch (maybe even hardcoded as done before).

PR #312, opened 2022-03-23
sys-fs/cryptsetup: move to ::coreos-overlay

tormath1:
this is required to enable fips mode on cryptsetup for stable release.

<hr>

changelog entry will be done on ::coreos-overlay.


pothos:
Since we only backport something that will be present when updating to a newer version I think it would be nice to have this stay in portage-stable. The move makes it harder to cherry-pick in the future.

tormath1:
> Since we only backport something that will be present when updating to a newer version I think it would be nice to have this stay in portage-stable. The move makes it harder to cherry-pick in the future.
> 
> Downstream patches that are not upstreamed to Gentoo would be lost on update but here this is probably not the case anymore because the newer versions have the flag.

@pothos I think I understand your point - but why would we need to cherry-pick this commits ? To me, once in Stable it should not have to move ?
Since this commits will be on `flatcar-3033` they should not interfere with future versions of cryptsetup.

pothos:
In the past there were cases where we updated software on the stable branch and for the LTS branch it is even a bit more likely that this can happen. Anyway, I just thought it can spare us some work and make diffing easier because I don't really see a benefit in the move.

tormath1:
@pothos it was mostly to stay consistent with the concept of: flatcar modification -> ::coreos-overlay. Once the CI done, I'll move back cryptsetup to this overlay and simply apply the following commit on it: https://github.com/flatcar-linux/coreos-overlay/pull/1749/commits/5495dbfb7e2d65a6ce2006b2706fa93e2a075c2e

krnowak (approved):

PR #313, opened 2022-03-24
Move virtual/rust to coreos-overlay and drop the rust workflow

krnowak:
Rust has two packages - virtual/rust and dev-lang/rust. The former is in portage-stable, the latter in coreos-overlay. We are not really following gentoo in rust updates - we have our own automation that updates those packages. Portage-stable is not a correct place for packages modified by us, thus move virtual/rust to coreos-overlay.

Eventually we will want to follow gentoo in the rust updates, so maybe at some point we will move both of the packages to portage-stable. But for now, we have made some modifications to dev-lang/rust package. Whether these modifications are still necessary or could be upstreamed is still to be seen.

Should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1761.

CI: http://localhost:9091/job/os/job/manifest/5241/cldsv/


krnowak:
CI passed.

jepio (approved):

dongsupark (approved):
OK if Github Actions tests also passed.

PR #318, opened 2022-03-30
sec-policy/selinux-sssd: Downgrade to match other SELinux packages

krnowak:
It was my mistake during the review to allow adding a package that is
much newer than the other SELinux packages we currently have. And for
some reason I started getting build failures when updating packages
from 2020:

```
!!! All ebuilds that could satisfy ">=sec-policy/selinux-base-policy-2.20210908-r1" for /build/amd64-usr/ have been masked.
!!! One of the following masked packages is required to complete your request:
- sec-policy/selinux-base-policy-9999::coreos (masked by: missing keyword)

(dependency required by "sec-policy/selinux-sssd-2.20210908-r1::portage-stable" [ebuild])
(dependency required by "sys-auth/sssd-2.3.1-r4::coreos[selinux]" [ebuild])
(dependency required by "app-admin/sudo-1.9.5_p2::coreos[sssd]" [ebuild])
(dependency required by "coreos-base/coreos-0.0.1-r302::coreos" [ebuild])
(dependency required by "coreos-base/coreos-dev-0.1.0-r81::coreos" [ebuild])
(dependency required by "coreos-devel/board-packages-0.0.1-r9::coreos" [ebuild])
(dependency required by "coreos-devel/board-packages" [argument])
```

It's from gentoo commit 1f7a3661f47b4da14ac3ad8c85f54f6b144104e4.

CC: @JAORMX 

CI: It is a part of package updates, 2020 edition - http://localhost:9091/job/os/job/manifest/5298/cldsv/


krnowak:
It is not a downgrade per-se - the same version is now built as before this change. What happened is that I synced selinux-sssd with an older version of gentoo repo, which didn't stabilize the newer versions. I don't know why I started getting those failures whereas nightlies were just fine, but still better to keep all the SELinux packages at the same version.

jepio (approved):
LGTM. Alternative would have been to add a temporary mask to the profiles in coreos-overlay

JAORMX (approved):
oops! thanks for fixing this!

krnowak:
> LGTM. Alternative would have been to add a temporary mask to the profiles in coreos-overlay

Right, next time I'll probably do it, but I rather hope that there won't be "next time".

> oops! thanks for fixing this!

No worries. :)

PR #321, opened 2022-04-11
dev-python/distro: Import from gentoo

jepio:
# dev-python/distro: Import from gentoo

This package is needed by the current version of google-compute-engine that we
ship but the dependency has been missing. We haven't noticed because the
package has actually been broken since python2 was dropped from the tree.
distro is needed to replace some functionality removed from the python standard
library around python 3.7.

## How to use

Can be tested together with the coreos-overlay PR (https://github.com/flatcar-linux/coreos-overlay/pull/1802), but merged independently (as long as before).

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]


tormath1 (approved):

tormath1:
Thanks - just edited the PR description to link the coreos-overlay's PR :)

jepio:
Thank you @tormath1

krnowak:
@jepio : Is that a OEM package? I think that dev-python/distro-oem in coreos-overlay should address this?

krnowak:
(By "this" I meant the google-compute-engine stuff.)

jepio:
It is not (quite) an oem package, the GCP oem package builds a container image that has normal `dev-lang/python` and needs the normal dev-python/distro.

krnowak:
Hm, GCP oem package probably should be using in dev-lang/python-oem package instead, but eh. If we can actually replace `dev-lang/{python,distro}-oem` with `dev-lang/{python,distro}` at some point, then it would be a win for maintenance.

jepio:
GCP needs a full blown python install so python-oem is not enough.
I have a different PR I will be opening to propose shipping a pre-built `dev-lang/python`, which is needed for things like azure extensions.

One way to completely drop python-oem and distro-oem is if we adopt the oem-as-sysext approach, then we could use stock distro/python packages.

PR #323, opened 2022-04-13
Package updates, 2021 edition, part 1 - fallout fixes from the update of profiles in portage-stable

krnowak:
The update of the profiles in portage-stable dropped support for freebsd kernels, masked sys-libs/e2fsprogs-libs (in favor of a newer version of sys-fs/e2fsprogs package) and masked new versions of dev-vcs/git. This means that some use flags were dropped or packages got masked, so packages that were using them also needed an update.

The reason for updating profiles is to unblock an update of setuptools, that will come in the followup PR.

Should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1819.

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5625/cldsv/

- app-crypt/mit-krb5
  - 1.19.2-r2 -> 1.19.2-r4
  - drops sys-libs/e2fsprogs-libs dep in favor of e2fsprogs
- app-emulation/qemu
  - 6.1.0-r1 -> 7.0.0-r1
  - stable for amd64, so updated keywords for arm64
  - IUSE gnutls became +gnutls; keep as is (now required by vnc which
    we enable)
  - IUSE kernel_linux and kernel_FreeBSD dropped (seems like this is
    implicit or something), xfs also dropped (we didn't enable it)
  - new IUSE pam, which we don't enable
  - deps changed for infiniband, which we don't enable
- dev-util/catalyst
  - still at 3.0.18
  - IUSE kernel_linux dropped
- net-nds/openldap
  - 2.4.58 -> 2.4.58-r2
  - got stable on both amd64 and arm64, updated keywords in overlay
  - IUSE libressl dropped, does not matter for us
  - trying to drop the openldap fixes from coreos/config in overlay
  - IUSE -cxx changed to cxx, but that's no behaviour change - it's
    still disabled by default
  - new dep on virtual/libcrypt if crypt USE flag is enabled, which we
    don't do
  - changed dep from sys-libs/e2fsprogs-libs to sys-fs/e2fsprogs -
    Gentoo is trying to drop the former
- sys-fs/e2fsprogs
  - 1.46.4 -> 1.46.5-r1
  - drops dep on sys-libs/e2fsprogs-libs
  - new IUSE test, which we don't enable
  - new IUSE tools, which we leave as enabled by default - nothing
    should change, previously the tools were built unconditionally
- sys-libs/e2fsprogs-libs
  - removed, dead package
- sys-libs/ldb
  - 2.3.0-r2 -> 2.4.1-r1
  - some deps bumped (sys-libs/tevent, sys-libs/tdb, sys-libs/talloc)
  - moved back to portage-stable
- sys-libs/talloc
  - 2.3.2 -> 2.3.3
  - got stable on amd64 and arm64, so we updated keywords in overlay
  - new version of samba or ldb in overlay needs it
- sys-libs/tdb
  - still at 1.4.5
  - no changes actually
- sys-libs/tevent
  - 0.10.2 -> 0.11.0
  - IUSE elibc_glibc dropped (it's implicit now)
  - new version of samba or ldb in overlay needs it


app-emulation/qemu/qemu-6.2.0.ebuild
```diff
@@ -33,9 +33,9 @@ LICENSE="GPL-2 LGPL-2 BSD-2"
 SLOT="0"
 
 IUSE="accessibility +aio alsa bpf bzip2 capstone +caps +curl debug +doc
-	+fdt fuse glusterfs gnutls gtk infiniband iscsi io-uring
-	jack jemalloc +jpeg kernel_linux
-	kernel_FreeBSD lzo multipath
+	+fdt fuse glusterfs +gnutls gtk infiniband iscsi io-uring
```
dongsupark, line 36: Starting from 6.2.0, because of the change to `+gnutls`, qemu pulls in gnutls by default, and that results in pulling in libtasn1 as well as nettle in the SDK. I am not sure I like the change, as it makes the SDK size bigger again.
However, as you wrote in the description, the `vnc` requires gnutls. If we would want to disable gnutls, we would need to disable vnc as well. In some cases vnc would be needed for debugging, so I think we should keep enabling both vnc and gnutls.

dongsupark (approved):
How about adding a new entry for qemu in `changelog/security/`, like below:
```
- SDK: qemu ([CVE-2021-20203](https://nvd.nist.gov/vuln/detail/CVE-2021-20203), [CVE-2021-3713](https://nvd.nist.gov/vuln/detail/CVE-2021-3713), [CVE-2021-3930](https://nvd.nist.gov/vuln/detail/CVE-2021-3930), [CVE-2021-3947](https://nvd.nist.gov/vuln/detail/CVE-2021-3947), [CVE-2021-4145](https://nvd.nist.gov/vuln/detail/CVE-2021-4145))
```

app-emulation/qemu/qemu-6.2.0.ebuild
krnowak: I'm not worried about the size of SDK as much as the size of stuff installed into the boot partition in the production image for example. I mean, yes, it's annoying the SDK gets bigger, but if the SDK size is a problem, then I'm sure we have some unnecessary stuff there that could be dropped right away.

krnowak:
> How about adding a new entry for qemu in `changelog/security/`, like below:
> 
> ```
> - SDK: qemu ([CVE-2021-20203](https://nvd.nist.gov/vuln/detail/CVE-2021-20203), [CVE-2021-3713](https://nvd.nist.gov/vuln/detail/CVE-2021-3713), [CVE-2021-3930](https://nvd.nist.gov/vuln/detail/CVE-2021-3930), [CVE-2021-3947](https://nvd.nist.gov/vuln/detail/CVE-2021-3947), [CVE-2021-4145](https://nvd.nist.gov/vuln/detail/CVE-2021-4145))
> ```

Thanks, I'm going to use it. But since the PR was quite old, I refreshed it and then we got an update to qemu 7.0.0. Any security updates on this one?

dongsupark:
> > How about adding a new entry for qemu in `changelog/security/`, like below:
> > ```
> > - SDK: qemu ([CVE-2021-20203](https://nvd.nist.gov/vuln/detail/CVE-2021-20203), [CVE-2021-3713](https://nvd.nist.gov/vuln/detail/CVE-2021-3713), [CVE-2021-3930](https://nvd.nist.gov/vuln/detail/CVE-2021-3930), [CVE-2021-3947](https://nvd.nist.gov/vuln/detail/CVE-2021-3947), [CVE-2021-4145](https://nvd.nist.gov/vuln/detail/CVE-2021-4145))
> > ```
> 
> Thanks, I'm going to use it. But since the PR was quite old, I refreshed it and then we got an update to qemu 7.0.0. Any security updates on this one?

In that case, can you please 2 more CVEs like below?
```
[CVE-2022-26353](https://nvd.nist.gov/vuln/detail/CVE-2022-26353), [CVE-2022-26354](https://nvd.nist.gov/vuln/detail/CVE-2022-26354)
```

krnowak:
> In that case, can you please 2 more CVEs like below?
> 
> ```
> [CVE-2022-26353](https://nvd.nist.gov/vuln/detail/CVE-2022-26353), [CVE-2022-26354](https://nvd.nist.gov/vuln/detail/CVE-2022-26354)
> ```

Thanks, added.

dongsupark (approved):

PR #329, opened 2022-05-18
Package updates, 2021 edition, part 2 - python update

krnowak:
Python updated turned out to be quite a large update, mostly because Gentoo did an overhaul of the way python packages can be built and the overhaul became stable. It is done through `distutils-r1.eclass`, and it started pulling a bunch of packages. Those in turn had a lot of dependencies which in effect ballooned this PR. But, in essence, this PR is about updating dev-lang/python and dev-python packages, and syncing the relevant eclasses (basically python eclasses + eclasses that these python eclasses also inherit).

Contents of this PR:

- dev-db/sqlite
  - 3.34.1 -> 3.38.1
  - JSON extension is not build, because it's built-in now
- dev-lang/python
  - 3.9.8 -> 3.9.12
  - new IUSE flags which we don't enable - lto, pgo
- dev-python/cython
  - 0.29.23 -> 0.29.28
  - use EAPI 8
  - use new python build infra (setuptools)
- dev-python/docutils
  - 0.17.1 -> 0.18.1
  - EAPI 8
- dev-python/flit_core
  - new package - 3.7.1
  - dependency of distutils eclass
- dev-python/gpep517
  - new package - 3
  - pulls in dev-python/installer
  - dependency of distutils eclass
- dev-python/importlib_metadata
  - new package - 4.11.3
  - dependency of dev-python/setuptools and dev-python/nspektr
  - pulls in dev-python/zipp
- dev-python/installer
  - new package - 0.5.0
  - dependency of dev-python/gpep517
- dev-python/jaraco-context
  - new package - 4.1.1-r2
  - dependency of dev-python/jaraco-text and dev-python/nspektr
- dev-python/jaraco-functools
  - new package - 3.5.0-r2
  - dependency of dev-python/jaraco-text and dev-python/nspektr
  - pulls in dev-python/more-itertools
- dev-python/jaraco-text
  - new package - 3.7.0-r2
  - dependency of dev-python/setuptools
  - pulls in dev-python/jaraco-context, dev-python/jaraco-functools
- dev-python/jinja
  - 2.11.2-r1 -> 3.1.1
  - use EAPI 8
  - use new python build infra (setuptools)
- dev-python/lazy-object-proxy
  - 1.5.0 -> 1.7.1
  - use EAPI 8
- dev-python/lxml
  - 4.6.3-r1 -> 4.8.0-r1
  - use EAPI 8
  - use new python build infra (setuptools)
- dev-python/mako
  - 1.1.4 -> 1.2.0
[...]

PR #331, opened 2022-05-19
dev-python: Drop pylint and its dependencies

krnowak:
It does not seem to be a build dependency of python.

This is an experiment. I'm trying to see if I can drop a one possibly unused python package (and its dependencies), so https://github.com/flatcar-linux/portage-stable/pull/329 could be smaller.

Should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1909.

CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5676/

~- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)~
Changelog not needed - this PR just drops packages from SDK.


dongsupark (approved):
Thanks

PR #332, opened 2022-05-23
sys-power/acpid: new package

tormath1:

<hr>

- [ ] https://github.com/flatcar-linux/coreos-overlay/pull/1911


krnowak:
The package puts some files into `/etc`. Do we want them to be in the image? Maybe we should add `/etc/acpi` into `INSTALL_MASK`?

krnowak:
> > The package puts some files into `/etc`. Do we want them to be in the image? Maybe we should add `/etc/acpi` into `INSTALL_MASK`?
> 
> Yeah - I think it's worth trying that. `INSTALL_MASK` can be defined directly in the ebuild or do we need to add it to `::coreos-overlay` profile ?

Should go to the overlay, really. Such a change can easily get lost. The only thing I'm wondering about is whether this should go to a per-package env override (like [here](https://github.com/flatcar-linux/coreos-overlay/blob/main/coreos/config/env/sys-boot/grub)) or into a profile. I suppose if no other package puts stuff into `/etc/acpi` then INSTALL_MASK could go into profile. Otherwise, per-package env override.

tormath1:
> > > The package puts some files into `/etc`. Do we want them to be in the image? Maybe we should add `/etc/acpi` into `INSTALL_MASK`?
> > 
> > 
> > Yeah - I think it's worth trying that. `INSTALL_MASK` can be defined directly in the ebuild or do we need to add it to `::coreos-overlay` profile ?
> 
> Should go to the overlay, really. Such a change can easily get lost. The only thing I'm wondering about is whether this should go to a per-package env override (like [here](https://github.com/flatcar-linux/coreos-overlay/blob/main/coreos/config/env/sys-boot/grub)) or into a profile. I suppose if no other package puts stuff into `/etc/acpi` then INSTALL_MASK could go into profile. Otherwise, per-package env override.

Ok done - as expected the daemon is failing if no configuration is provided into `/etc/acpi/events`. We should provide documentation with an Ignition config to show how to correctly configure and start the `acpi` daemon.

krnowak:
@tormath1: I assume that this daemon is not started at boot by default. So yeah, docs would be useful.

krnowak (approved):

tormath1:
> @tormath1: I assume that this daemon is not started at boot by default. So yeah, docs would be useful.

@krnowak thanks for the review - PR's doc: https://github.com/flatcar-linux/flatcar-docs/pull/247

PR #334, opened 2022-06-10
policy/sssd, libs/ding , liburing: add ARM64 support

t-lo:
This change bumps liburing-2.1 to upstream -rc2 and ingests a cross compiler fix (https://github.com/gentoo/gentoo/pull/24733).

Also, it removes the mask for ARM64 in sec-policy/selinux-sssd to match Gentoo upstream.

## How to use

Build the image
```
./run_sdk_container -t 
   ./ build_packages --board=arm64-usr
   ./build_image --board=arm64-usr
   ./image_to_vm.sh ...
```

Launch the image
```
./flatcar_production_qemu_uefi.sh
[...]
core@localhost ~ $ sssd --version
2.3.1
```

## Testing done

Local arm64 test build succeeded, image boots, sssd prints version.

No changelog required since changelog entry will be added to PR in coreos-overlay.

### DIFF
--- a/sec-policy/selinux-sssd/selinux-sssd-2.20200818-r2.ebuild
+++ b/sec-policy/selinux-sssd/selinux-sssd-2.20200818-r2.ebuild
@@ -11,5 +11,5 @@ inherit selinux-policy-2
 DESCRIPTION="SELinux policy for sssd"
 
 if [[ ${PV} != 9999* ]] ; then
-	KEYWORDS="amd64 -arm ~arm64 ~mips x86"
+	KEYWORDS="amd64 -arm arm64 ~mips x86"
 fi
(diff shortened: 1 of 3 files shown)


jepio:
We prefer to do keywording of gentoo ebuilds by adding keywords in the profile in coreos-overlay, so as not to deviate from gentoo. If this works we can even make a stabilization request on gentoo's bugzilla.

That would leave only the liburing part of the commit, which looks OK.

t-lo:
@jepio using the profile to override masks would only apply to ding-libs as upstream's sec-policy/selinux-sssd is unmasked for ARM64. I believe we added the mask ourselves. So I'll revert the unmask for ding-libs and leave selinux-sssd unmasked to match upstream state.

Also, I'll mask AMD64 and ARM64 in liburing-2.1-rc2 since that's what upstream does, and unmask it in coreos-overlay's profile.

t-lo:
Done, PR and summary updated.

jepio (approved):
OK. Running CI

jepio:
CI looks good here too.

PR #335, opened 2022-06-14
app-crypt/adcli: move adcli from coreos-overlay

dongsupark:
To fix build issues that happen in adcli 0.9 with glibc 2.34, we should sync adcli with upstream Gentoo, where the build issue is already fixed.

As Gentoo has the ebuild under the category `app-crypt`, we simply move from adcli from coreos-overlay to portage-stable, move adcli to the app-crypt category, and update the version to 0.9.1-r2.

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/1957.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5869/cldsv/


dongsupark:
Please see https://github.com/flatcar-linux/portage-stable/pull/340 instead.

PR #350, opened 2022-08-18
dev-libs/cyrus-sasl: Move from coreos-overlay and update

krnowak:
It's from Gentoo commit 728228559d150db1a238accfa71f9eee7ae9bcd1.

This used to be in coreos-overlay, but upstream already has all the
patches we were carrying. The cross-compilation issue will be
addressed with an environment variable override in overlay.

Should be merged with https://github.com/flatcar-linux/coreos-overlay/pull/2096.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/173/cldsv/


krnowak:
CI passed.

jepio (approved):

PR #355, opened 2022-09-06
dev-lang/duktape: Sync with Gentoo

krnowak:
It's from Gentoo commit 8c0df49ac77e2cc9fd33a11d99db49e1931c11a9.

Not an update really - old ebuild got dropped and the unstable ebuild we used so far became stable.

Should be merged with COREOS_OVERLAY_PR.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/225/cldsv


krnowak:
CI passed (more or less - cilium base failed, probably because the tests were using older mantle docker container, without the fix for cilium).

tormath1 (approved):
Good! It became stable because we submitted a stabilization request a few months ago: https://bugs.gentoo.org/834909

PR #363, opened 2022-09-23
.github: Add an automatic package updates workflow

krnowak:
It will run on every Monday and will try to update packages listed in the package listing. This can be used to gradually add more and more packages under the automation.

## Testing done

Ran on my own fork - added app-arch/zstd and sys-devel/gcc to the listing and run the workflow there. The resulting PR is here: https://github.com/krnowak/portage-stable/pull/2

### DIFF
--- a/.github/workflows/packages-list
+++ b/.github/workflows/packages-list
@@ -1,0 +1,1 @@
+# Please keep the list sorted!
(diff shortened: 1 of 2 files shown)


pothos (approved):
Looks good!

PR #367, opened 2022-09-30
Update profiles, licenses and scripts

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/331/cldsv

- relevant changes:
  - hardened profile: -berkdb -gdbm -fortran -cli -dri dropped from USE flags. We have -berkdb base/make.defaults. Nothing changed in builds wrt -gdbm.

- semi-relevant changes:
  - amd64: The stack-realign USE flag is masked for sys-libs/glibc and sys-libs/ncurses. It is about fixing a x32/x86 problem, so not relevant for us. Also, our versions are too old, and don't have this USE flag yet.
  - all: The hostnamed-fallback USE flag of sys-apps/systemd is not masked any more. It is about being able to change hostname without polkit. We didn't use it anyway.
  - arm64: Some masks for net-fs/samba USE flags were dropepd (ads addc python). We didn't enable them anyway.
  - arm64: The apparmor USE flag is unmasked. We didn't enable it anyway.
  - all: Switched to python 3.10 by default. We should move too, eventually, but probably first move to virtual/libcrypt-2. Currently using 3.9.
  - all: The llvm-libunwind and miri USE flags are masked for dev-lang/rust. So our ebuild likely needs some sync with upstream, because ours doesn't have llvm-libunwind. The miri flag is not enabled anyway.
  - all: The symlink USE flag is masked for app-arch/{pigz,pbzip2,lbzip2}. This USE flag was creating a symlink like gzip -> pigz or bzcat -> {p,l}bzcat. We don't enable this flag anyway.
  - all: The gprofng USE flag masked for sys-devel/binutils. Does not seem to be used by the ebuilds anway.
  - all: The custom-cflags USE flag is masked for sys-devel/gcc. We didn't use it anyway.
  - all: The new merged-usr profile added. It's probably something we could use for SDK, but not for production images.
  - all: bash 5.2 and readline 8.2 masked for now. It's undergoing testing.
  - all: ~x11-drivers/nvidia-drivers-515.76 masked (causes issues for RTX30xx users). Not yet relevant for us, since we are still at 510.73, but something to keep in mind when updating.
  - all: =dev-util/pahole-1.24 masked (breaks kernel builds). We are still using 1.20 (or 1.23 if my dev-util packages update gets merged).
  - all: =sys-apps/gawk-5.2.0 masked (crashes on double free). We are still on 5.1.0.

There isn't anything changeloggable.

Nothing stands out.


krnowak:
CI passed.

scripts/bootstrap.sh
```diff
@@ -169,8 +169,8 @@ cleanup() {
 }
 
 pycmd() {
-	[[ ${DEBUG} = "1" ]] && echo /usr/bin/python -c "$@" > /dev/stderr
-	/usr/bin/python -c "$@"
+	[[ ${DEBUG} = "1" ]] && echo /usr/bin/python3 -c "$@" > /dev/stderr
+	/usr/bin/python3 -c "$@"
```
dongsupark, line 173: Do we still need this python change for Flatcar?
Also it is hard to know it is a Flatcar patch, as it does not have comments, only commit messages.

krnowak: I suppose we can try dropping this. This should work if we do the bootstrap from the newish version of Flatcar (so stable or newer, as LTS 2022 still has python2 as the default).

krnowak: Dropped python3 modification. I'll rerun the CI to see if it works.

dongsupark (approved):

PR #369, opened 2022-10-10
app-editors/vim: Sync with Gentoo

krnowak:
Updates vim to 9.0.0655.

Needs to be merged with https://github.com/flatcar/coreos-overlay/pull/2210.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/322/cldsv

Nothing interesting changed.


krnowak:
CI passed.

krnowak:
> The code looks good. However, `packages_all_arches/502` of the CI is now gone, because that ran 10 days ago unfortunately. Anyway I do not expect any regression from the vim update.

Things were all green if I remember correctly, otherwise I would have noted it in the PR.

Thanks for the review.

PR #379, opened 2022-10-26
.github: print date of today in the PR title

dongsupark:
Let's add date of today in the PR title, to avoid confusion when seeing the same title for multiple different weekly update PRs.

### DIFF
--- a/.github/workflows/update-packages-from-list.yml
+++ b/.github/workflows/update-packages-from-list.yml
@@ -62,17 +62,19 @@ jobs:
               updated=1
               count=$(git -C portage-stable rev-list --count "${old_head}..${new_head}")
           fi
+          todaydate=$(date +%Y-%m-%d)
           echo "UPDATED=${updated}" >>"${GITHUB_OUTPUT}"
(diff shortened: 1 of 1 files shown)


dongsupark:
Tested via my fork. https://github.com/dongsupark/portage-stable/pull/11
Works well.

.github/workflows/update-packages-from-list.yml
```diff
@@ -73,6 +75,6 @@ jobs:
           branch: buildbot/automatic-package-updates
```
krnowak, line 75: Question that I have here is what would be the desired outcome when the weekly package updates PR is not handled within the week:
- create another PR next Monday (would need a differently named branch too)
- update the existing PR (stays as is)

What do you think?

dongsupark: I would prefer to have another PR for a new week.
If we would explicitly want to merge new updates into an existing PR, then we could simply close the existing one, and run new Github Actions.

dongsupark: Done

krnowak (approved):

PR #383, opened 2022-11-07
Fix dev-libs/openssl build in stable channel

krnowak:
Add edo.eclass from Gentoo.

It's from Gentoo commit 6ce8bc59bc81e8a509570c7364e7177bdc181b16.

The dev-libs/openssl package started to use the edo eclass, which was not available yet in stable channel. Add it.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/701/cldsv

- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

### DIFF
--- a/eclass/edo.eclass
+++ b/eclass/edo.eclass
@@ -1,0 +1,48 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: edo.eclass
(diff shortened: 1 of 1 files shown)


krnowak:
3227 stays at 3.0.3 with extra patches.

PR #385, opened 2022-11-15
.github: add monthly workflow update-metadata-glsa

dongsupark:
Add a new monthly workflow `update-metadata-glsa`, to sync GLSA(Gentoo Linux Security Advisories) with upstream Gentoo.

While most individual metadata xml files are available on the upstream [mirror](https://github.com/gentoo-mirror/gentoo), its binary data as well as timestamp files do not exist on the public Git repo. So it is necessary to first rsync those data from `rsync.gentoo.org`, before actually running `sync-with-gentoo` script.

Remove unnecessary options in `update-packages-from-list`. As we already merged necessary changes for `sync-with-gentoo` into master branch of the flatcar-build-scripts repo, it is not necessary any more to point to `krnowak/stuff` branch. Remove it.

The checkout action of the Github Actions defaults to `fetch-depth=1`, when no fetch-depth is given. So it is better for efficiency not to specify fetch-depth.

## Testing done

Tested from my own [fork](https://github.com/dongsupark/portage-stable).

### DIFF
--- a/.github/workflows/update-metadata-glsa.yml
+++ b/.github/workflows/update-metadata-glsa.yml
@@ -1,0 +1,32 @@
+name: Keep GLSA metadata updated
+on:
+  schedule:
+    - cron:  '0 7 1 * *'
+  workflow_dispatch:
+
+jobs:
+  keep-glsa-metadata-updated:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout portage-stable
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update GLSA metadata
+        id: update-glsa-metadata
+        run: |
+          rm -rf metadata/glsa
+          mkdir --parents metadata/glsa
+          rsync --archive rsync://rsync.gentoo.org/gentoo-portage/metadata/glsa/* ./metadata/glsa/
+          todaydate=$(date +%Y-%m-%d)
+          echo "TODAYDATE=${todaydate}" >>"${GITHUB_OUTPUT}"
+      - name: Create pull request for main branch
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: buildbot/monthly-glsa-metadata-updates-${{steps.update-glsa-metadata.outputs.TODAYDATE }}
+          delete-branch: true
+          base: main
+          title: Monthly GLSA metadata ${{steps.update-glsa-metadata.outputs.TODAYDATE }}
+          body: Updated GLSA metadata
+          labels: main
--- a/.github/workflows/update-packages-from-list.yml
+++ b/.github/workflows/update-packages-from-list.yml
@@ -20,13 +20,14 @@ jobs:
           path: gentoo
           # Gentoo is quite a large repo, so limit ourselves to last
           # quarter milion of commits. It is about two years worth of changes.
+          # Is is needed by the sync script to find out the hash of the last commit
+          # that made the changes to the package.
           fetch-depth: 250000
           ref: master
       - name: Checkout build scripts
         uses: actions/checkout@v3
         with:
           repository: flatcar/flatcar-build-scripts
-          ref: krnowak/stuff
           path: flatcar-build-scripts
       - name: Update listed packages
         id: update-listed-packages


.github/workflows/update-metadata-glsa.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          path: ./portage-stable
+      - name: Checkout Gentoo
+        uses: actions/checkout@v3
+        with:
+          repository: gentoo-mirror/gentoo
+          path: gentoo
+          fetch-depth: 1
```
jepio, line 21: you mentioned this is the default, is it needed here?

.github/workflows/update-metadata-glsa.yml
```diff
@@ -0,0 +1,58 @@
...
+          updated=0
+          if [[ "${new_head}" != "${old_head}" ]]; then
+              updated=1
+          fi
+          todaydate=$(date +%Y-%m-%d)
+          echo "UPDATED=${updated}" >>"${GITHUB_OUTPUT}"
+          echo "TODAYDATE=${todaydate}" >>"${GITHUB_OUTPUT}"
+      - name: Create pull request for main branch
+        uses: peter-evans/create-pull-request@v4
+        if: steps.update-glsa-metadata.outputs.UPDATED == 1
```
jepio, line 49: i found the following in the action documentation:

> How the action behaves:
> 
> If there are changes (i.e. a diff exists with the checked-out base branch), the changes will be pushed to a new branch and a pull request created.
If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.
If a pull request already exists it will be updated if necessary. Local changes in the Actions workspace, or changes on the base branch, can cause an update. If no update is required the action exits silently.
If a pull request exists and new changes on the base branch make the pull request unnecessary (i.e. there is no longer a diff between the pull request branch and the base), the pull request is automatically closed. Additionally, if delete-branch is set to true the branch will be deleted.

So we could use these properties and get rid of the `UPDATED` checks/rev-parsing.

.github/workflows/update-packages-from-list.yml
```diff
@@ -18,15 +18,11 @@ jobs:
         with:
           repository: gentoo/gentoo
           path: gentoo
-          # Gentoo is quite a large repo, so limit ourselves to last
-          # quarter milion of commits. It is about two years worth of changes.
-          fetch-depth: 250000
```
krnowak, line 23: Please keep it - it's needed by the sync script to find out the hash of the last commit that made the changes to the package in Gentoo. I think that the number can be lowered at some point, but certainly not to just 1.

.github/workflows/update-metadata-glsa.yml
```diff
@@ -0,0 +1,58 @@
+name: Keep GLSA metadata updated
+on:
+  schedule:
+    - cron:  '0 7 1 * 1'
```
krnowak, line 4: This means do perform the action [at 7 in the morning on the first day of a month and on Monday](https://crontab.guru/#0_7_1_*_1). I have no idea if this means that this will be executed only when first day of a month is Monday at the same time, or it will be executed twice when first day of a month is not Monday (once on the first day, second time on Monday). Or scheduler will be smart and execute it on the first Monday after first day of a month…

.github/workflows/update-metadata-glsa.yml
```diff
@@ -0,0 +1,58 @@
...
+          GENTOO_REPO=../gentoo ../flatcar-build-scripts/sync-with-gentoo "metadata/glsa"
+          cd ..
+          new_head=$(git -C portage-stable rev-parse HEAD)
+          updated=0
+          if [[ "${new_head}" != "${old_head}" ]]; then
+              updated=1
+          fi
+          todaydate=$(date +%Y-%m-%d)
+          echo "UPDATED=${updated}" >>"${GITHUB_OUTPUT}"
+          echo "TODAYDATE=${todaydate}" >>"${GITHUB_OUTPUT}"
```
krnowak, line 46: I'm wondering if we even need this. The value of sync-with-gentoo is that it will also find the last commit in Gentoo repo that modified the synced package. GLSA stuff comes from outside the repo, so it won't be a part of the git history. I suppose you could do:

```suggestion
          cd portage-stable # probably won't be needed if we only clone portage-stable above
          rm -rf metadata/glsa # not sure if this is needed too, rsync maybe will remove old stuff
          # need to rsync binary data Manifest* and timestamp* which are not in the Git repos.
          rsync --archive rsync://rsync.gentoo.org/gentoo-portage/metadata/glsa/{Manifest,timestamp}* ./metadata/glsa/
          todaydate=$(date +%Y-%m-%d)
          echo "TODAYDATE=${todaydate}" >>"${GITHUB_OUTPUT}"
```

That way, you wouldn't even need to clone the Gentoo repo. You could keep the clone of portage-stable as a current working directory.

.github/workflows/update-metadata-glsa.yml
dongsupark: removed fetch-depth

.github/workflows/update-metadata-glsa.yml
dongsupark: removed UPDATED

.github/workflows/update-packages-from-list.yml
dongsupark: Brought back fetch-depth

.github/workflows/update-metadata-glsa.yml
dongsupark: fixed the crontab

.github/workflows/update-metadata-glsa.yml
dongsupark: simplified with some more changes

.github/workflows/update-packages-from-list.yml
```diff
@@ -20,13 +20,14 @@ jobs:
           path: gentoo
           # Gentoo is quite a large repo, so limit ourselves to last
           # quarter milion of commits. It is about two years worth of changes.
+          # Is is needed the sync script to find out the hash of the last commit
```
krnowak, line 23:
```suggestion
          # Is is needed by the sync script to find out the hash of the last commit
```

krnowak (approved):

dongsupark:
Thanks for reviews!

PR #398, opened 2022-12-09
sys-apps/portage: Move from coreos-overlay and update

krnowak:
Should be merged together with https://github.com/flatcar/coreos-overlay/pull/2325 and https://github.com/flatcar/scripts/pull/606.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/513/cldsv/

This also puts sys-apps/portage to the list of automatically updated packages, so it will be kept up-to-date.

Bringing it here, since I think we don't need our custom patches in portage. We can do without them. Also updating it, because I'd like to update and patch gentoolkit and it needs a newer portage version than the one we had so far.

--

- sys-apps/portage: (DEV)
  - from 3.0.28-r2 to 3.0.41-r2
  - added dependency on sys-apps/coreutils and sys-apps/file
  - stopped building native extensions when cross-compiling
  - release notes: https://gitweb.gentoo.org/proj/portage.git/tree/NEWS?h=portage-3.0.41

- sys-apps/file: (DEV)
  - from 5.40-r3 to 5.43-r2
  - EAPI 8
  - release notes: https://mailman.astron.com/pipermail/file/2022-September/000857.html

- sys-libs/libseccomp: (PROD) (DEV)
  - from 2.5.1-r1 to 2.5.4
  - new IUSE experimental-loong (not enabled for us)
  - bumped dependency on kernel headers to 5.15
  - dropped obsolete accept keywords from overlay
  - needed by updated sys-apps/file package, as it uses _SNR_faccessat2, which was unavailable in 2.5.1
  - release notes: https://github.com/seccomp/libseccomp/releases/tag/v2.5.4 https://github.com/seccomp/libseccomp/releases/tag/v2.5.2 https://github.com/seccomp/libseccomp/releases/tag/v2.5.3

--

- [x] changelog
- [x] image diff


krnowak:
CI passed.

krnowak:
Did a package refresh and a CI rerunning, all green.

changelog/updates/2022-12-12-portage-update.md
```diff
@@ -0,0 +1,3 @@
+- portage ([3.0.41](https://gitweb.gentoo.org/proj/portage.git/tree/NEWS?h=portage-3.0.41))
+- file ([5.43](https://mailman.astron.com/pipermail/file/2022-September/000857.html))
```
dongsupark, line 2: Maybe add prefix `SDK: ` to both portage and file?

dongsupark (approved):
Looks good.
See below:

changelog/updates/2022-12-12-portage-update.md
krnowak: I tend to not add the entries for software that is only in SDK unless it's kind-of-a-major piece of software. In this case, both `portage` and `file` are also a part of the developer container (thus the `(DEV)` annotation in the description above), that's why I added the changelog without any prefixes.

We haven't yet written down what updates should go into the changelog, so my loosely rule so far was that I log updates of important software for production images and developer containers, and same for SDK with the `SDK: ` prefix.

dongsupark (approved):

PR #414, opened 2023-02-09
app-eselect/eselect-iptables: update to 20220320

dongsupark:
As `iptables` was updated to 1.8.8 in coreos-overlay, we need to update `eselect-iptables` to 20220320 as well, to avoid bugs having missing symlinks like `/sbin/ip6tables`, `/sbin/ip6tables-{save,restore}`.

See also https://cgit.gentoo.org/proj/eselect-iptables.git/commit/?id=2512407d3790.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/609/cldsv/

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) (not needed)


krnowak:
Could you please add this package to the packages list too in `.github` directory? Thanks!

krnowak (approved):

PR #431, opened 2023-03-21
Automate updates of system packages: app-arch

krnowak:
"System packages" means packages maintained by `base-system@gentoo.org`.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/723/cldsv

Should be merged together with https://github.com/flatcar/coreos-overlay/pull/2544.

--

- app-arch/rpm2targz:
  - still at 2021.03.16
  - updated keywords for other arches

- app-arch/sharutils:
  - still at 4.15.2-r2
  - updated keywords for other arches

- app-arch/zip: (PROD) (DEV)
  - from 3.0-r4 to 3.0-r7
  - EAPI 8
  - fixes for clang 15 and 16

- app-arch/zstd: (PROD) (DEV)
  - from 1.4.9 to 1.5.2-r3
  - EAPI 8
  - stable for both amd64 and arm64, so dropped accept keywords from overlay
  - dropped threads IUSE, it's now enabled by default
  - added new RDEPEND on sys-libs/zlib (it should have always been there)
  - release notes: https://github.com/facebook/zstd/releases/tag/v1.5.2 https://github.com/facebook/zstd/releases/tag/v1.5.1 https://github.com/facebook/zstd/releases/tag/v1.5.0

--


krnowak:
CI passed.

PR #432, opened 2023-03-23
Automate updates of system packages: app-*

krnowak:
"System packages" means packages maintained by `base-system@gentoo.org`.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/726/cldsv

--

- app-crypt/libmd: (PROD) (DEV)
  - still at 1.0.4
  - fixed configure check for some linker feature
  - updated metadata

- app-crypt/pinentry: (PROD) (DEV)
  - from 1.2.0 to 1.2.1-r1
  - EAPI 8
  - added verify-sig
  - moved app-eselect/eselect-pinentry dependency from DEPEND to IDEPEND
  - dropped dependency on sys-libs/libcap (but the caps IUSE is left intact, probably an omission)
  - dropped explicit setting of -std=gnu++11, current GCC uses this or newer standard by default
  - release notes: https://git.gnupg.org/cgi-bin/gitweb.cgi?p=pinentry.git;a=blob;f=NEWS;h=c080b34e57d01a6ccca9d2996d7096c42b1a3f84;hb=8ab1682e80a2b4185ee9ef66cbb44340245966fc

- app-editors/nano:
  - from 5.8-r2 to 7.2
  - EAPI 8
  - updated license to GPL-3+ LGPL-2.1+ || ( GPL-3+ FDL-1.2+ )
  - dropped split-usr IUSE, it was installing a symlink in /usr/bin/nano, pointing to /bin/nano; it added no value to the package
  - installs nanorc file with ebuild file highlighting rules

- app-misc/c_rehash: (PROD) (DEV)
  - still at 1.7-r1
  - updated keywords for other arches

- app-misc/editor-wrapper:
  - still at 4-r1
  - updated keywords for other arches

- app-portage/elt-patches:
  - from 20220831 to 20221210
  - adds a patch that allows -Werror and -Werror=* flags to pass through a filter at link time, so flags like -Werror=odr or -Werror=lto-type-mismatch can take effect

- app-text/manpager: (DEV)
  - still at 1
  - updated keywords for other arches

--


krnowak:
Will file a PR to scripts after BIg Merge.

